PR #74, opened 2014-06-08
Improved traversers

szymonm:
Added abstract class QueueBasedTraverser holds general logic for most
traversers.
Fixed errors in InfoKeepers.
Added InfoKeeper that uses IntToBoolean fastutil map to store Boolean
values.
Changed BreadthFirstTraverser so that first node is the node we start
traverse from (to be consistent with other traversals).

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
@@ -571,7 +571,7 @@ trait PathLengthTracker extends DepthFirstTraverser {
  * Finishing time of a node is time just after all `dir` neighbors of the node became finished
  * (or when it is visited if it has no `dir` neighbors that should be processed later).
  */
-trait DiscoveryAndDepthTimeTracker extends DepthFirstTraverser {
+trait DiscoveryAndFinishTimeTracker extends DepthFirstTraverser {
   private[this] var time: Int = _ // automatically initialized to 0 before first enqueue
 
   private lazy val discoveryTime = new IntInfoKeeper(false)
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
@@ -15,12 +15,10 @@ package com.twitter.cassovary.graph
 
 import com.twitter.cassovary.graph.GraphDir._
 import com.twitter.cassovary.graph.GraphUtils.RandomWalkParams
-import com.twitter.cassovary.graph.tourist.{IntInfoKeeper, PrevNbrCounter}
+import com.twitter.cassovary.graph.tourist.{BoolInfoKeeper, IntInfoKeeper, PrevNbrCounter}
 import com.twitter.logging.Logger
-import com.twitter.ostrich.stats.Stats
 import it.unimi.dsi.fastutil.ints.IntArrayFIFOQueue
-import it.unimi.dsi.fastutil.objects.ObjectArrayList
-import scala.collection.mutable
+import scala.annotation.tailrec
 import scala.util.Random
 
 /**
@@ -38,24 +36,32 @@ trait Traverser extends Iterator[Node] { self =>
  * Bounds an iterator to go no more than a specified maximum number of steps
  */
 trait BoundedIterator[T] extends Iterator[T] {
-  val maxSteps: Long
+  /**
+   * @return If option is defined, it is the maximal number of elements the iterator returns.
+   *         If is `None`, the iteratator is not bounded.
+   */
+  def maxSteps: Option[Long]
+
   private var numStepsTaken = 0L
 
-  abstract override def next = {
+  abstract override def next() = {
     numStepsTaken += 1
-    super.next
+    super.next()
   }
 
-  abstract override def hasNext = ( (numStepsTaken < maxSteps) && super.hasNext)
+  abstract override def hasNext = maxSteps match {
+    case Some(max) if numStepsTaken >= max => false
+    case _ => super.hasNext
+  }
 }
 
 /**
- * Randomly traverse the graph, going from one node to a random neighbor in direction {@code dir}.
+ * Randomly traverse the graph, going from one node to a random neighbor in direction `dir`.
  * @param graph the graph to traverse on
  * @param dir direction in which to traverse
  * @param homeNodeIds the ids of nodes that the traverser will go to next if the node has no
- * neighbors, or we reset the traversal with probability {@code resetProbability},
- * or if the number of out-going edges at the current node exceeds {@code maxNumEdgesThresh}
+ * neighbors, or we reset the traversal with probability `resetProbability`,
+ * or if the number of out-going edges at the current node exceeds `maxNumEdgesThresh`
  * @param resetProbability the probability of teleporting back to home node set at each step
  * @param maxNumEdgesThresh if set, do not traverse edges with > maxNumEdgesThresh outgoing edges
  * @param onlyOnce specifies whether the same node should only be allowed to be visited once
@@ -68,13 +74,12 @@ trait BoundedIterator[T] extends Iterator[T] {
 class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
                       resetProbability: Double, maxNumEdgesThresh: Option[Int], onlyOnce: Boolean,
                       randNumGen: Random, maxDepth: Option[Int], filterHomeNodeByNumEdges: Boolean)
-    extends Traverser {  self =>
+  extends Traverser {  self =>
 
   private var currNode: Node = null
-  private var homeNode: Node = null
   private val homeNodeIdSet = Set(homeNodeIds: _*)
 
-  private val seenNodesTracker = new IntInfoKeeper(self.onlyOnce)
+  private val seenNodesTracker = new BoolInfoKeeper(self.onlyOnce)
 
   protected def seenBefore(id: Int) = seenNodesTracker.infoOfNode(id).isDefined
   private var pathLength = 0
@@ -86,21 +91,21 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
 
   private def takeRandomStep(): Int = {
     val nextRandom = randNumGen.nextDouble()
-    val needToFilterByNumEdges = (filterHomeNodeByNumEdges || !(homeNodeIdSet contains currNode.id))
+    val needToFilterByNumEdges = filterHomeNodeByNumEdges || !(homeNodeIdSet contains currNode.id)
     if (nextRandom < resetProbability ||
-        (needToFilterByNumEdges && NodeUtils.hasTooManyEdges(dir, maxNumEdgesThresh)(currNode))) {
+      (needToFilterByNumEdges && NodeUtils.hasTooManyEdges(dir, maxNumEdgesThresh)(currNode))) {
       goHome()
     } else {
       currNode.randomNeighbor(dir, randNumGen).getOrElse(goHome())
     }
   }
 
-  def next = {
+  def next() = {
     val nextNodeId = if (currNode == null ||
-        (maxDepth.isDefined && pathLength >= maxDepth.get)) {
+      (maxDepth.isDefined && pathLength >= maxDepth.get)) {
       goHome()
     } else {
-      var randNextNodeId = takeRandomStep()
+      val randNextNodeId = takeRandomStep()
       if (onlyOnce && seenBefore(randNextNodeId)) {
         seenNodesTracker.clear()
         goHome()
@@ -108,7 +113,7 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
         randNextNodeId
       }
     }
-    seenNodesTracker.recordInfo(nextNodeId, 0)
+    seenNodesTracker.recordInfo(nextNodeId, false)
     pathLength += 1
     currNode = getExistingNodeById(graph, nextNodeId)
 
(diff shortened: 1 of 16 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/graph/tourist/IntInfoKeeper.scala
```diff
@@ -14,52 +14,23 @@
...
 
-  /**
-   * Record information {@code info} of node {@code id}.
-   */
-  def recordInfo(id: Int, info: Int) {
-    if (!(onlyOnce && infoPerNode.containsKey(id))) {
-      infoPerNode.put(id, info)
-    }
-  }
+  override protected def infoPerNode: mutable.Map[Int, Int] = underlyingMap.asInstanceOf[util.Map[Int, Int]]
```
szymonm: We still use underlyingMap in the VisitsCounter. I will leave it for now.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -38,24 +37,110 @@ trait Traverser extends Iterator[Node] { self =>
...
+  /**
+   * The priority of nodes adding to the queue.
+   */
+  def nodePriority: GraphTraverserNodePriority
+
+  /**
+   * Action performed when visiting node `node` (before iteratur retuns `node`).
+   * Should be implemented in subclass.
+   */
+  def visitNode(node: Node): Unit = {}
```
szymonm: The short notation will be deprecated in near future:
https://issues.scala-lang.org/browse/SI-7605
I suggest leaving it everywhere.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/tourist/PrevNbrCounter.scala
```diff
@@ -14,70 +14,70 @@
...
+  }
+
   /**
-   * Record the previous neighbor {@code nodeId} of {@code id}.
+   * Record the previous neighbor `nodeId` of `id`
    */
-  def recordInfo(id: Int, nodeId: Int) {
+  def recordPreviousNeighbor(id: Int, nodeId: Int) {
     if (!(onlyOnce && infoPerNode.containsKey(id))) {
       nbrCountsPerNodeOrDefault(id).add(nodeId, 1)
```
szymonm: In current fastutil (6.4.4) in Build.scala there is no addTo. Let's make it in a separate PR, ok? [93]

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+  
+  import NodeColor._
 
-  protected def visitPotentialNode(id: Int, info: T, dist: Int) {
-    val seen = seenBefore(id)
-    if (!(onlyOnce && seen)) {
-      storeForVisit(id, info)
+  val map = new BoolInfoKeeper(false)
+
+  def +=(kv: (Int, Color)) {
```
szymonm: Right, good point.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+  val Visited = Value
+}
 
-  protected def init(defaultInfo: T) {
-    homeNodeIds foreach { id => visitPotentialNode(id, defaultInfo, 0) }
-  }
+/**
+ * Stores information of three colors assigned to node ids (Ints).
+ *
+ * Initially every node has color `Unseen`.
```
szymonm: I would stay with Unenqueued. If we have Unseen, Enqueued it can confusing whether a node can be Seen and still not enqueued. In all our version we enqueue a node when we see it for the first time. Is this ok?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+ *
+ * Uses BoolInfoKeeper under the hood.
+ */
+trait WalkNodeColoring {
```
szymonm: I'm not sure but I think trait is more general as we can mix it to other class.

The cost is small - just {} when constructing. What do you think?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+}
+
+/**
+ * Traverses in DFS order (every node can be visited only once).
+ * @param graph the graph to traverse on
+ * @param dir direction in which to traverse
+ * @param homeNodeIds the ids of nodes that the walk starts from
+ * @param walkLimits limiting parameters
+ */
+final class DepthFirstTraverser(val graph: Graph, val dir: GraphDir, val homeNodeIds: Seq[Int],
```
szymonm: We need final only for @tailrec method findNextToVisit. Changed it.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+  override def limitAddedToQueue(nodes: Seq[Int]): Seq[Int] = nodes
+
+  override def shouldBeEnqueued(color: NodeColor.Color): Boolean = color != NodeColor.Visited
+}
+
+
+/**
+ * A queue based traverser trait that keeps track of visiting distance from `homeNodeIds`.
+ */
+trait PathLengthTracker extends QueueBasedTraverser {
```
szymonm: So let's leave PathLengthTracker and FinishingTimeTracker as optional when mixed in. Details coming soon.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -128,157 +218,227 @@ class RandomBoundedTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+trait DepthTracker extends QueueBasedTraverser with NodesSeenBefore {
+
+  private lazy val depthTracker = new IntInfoKeeper(true)
+
+  override protected def enqueue(nodes: Seq[Int], from: Option[Int]): Unit = {
+    val fromDepth = from.flatMap(depthTracker.infoOfNode).getOrElse(-1)
+    nodes foreach { id =>
+      depthTracker.recordInfo(id, fromDepth + 1)
+    }
+    super.enqueue(nodes, from)
```
szymonm: IIUC, depth and enqueue time are not the same.
For tree like:
  1
 /  \
2   3
    / 
   4
depth(4) = 2
enqueueTime(3) // so many nodes enqueued before

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
 
 /**
- * A traverser that keeps track of distance from homeNodeIds. If {@code onlyOnce} is true,
- * the same node is not traversed again.
+ * Optional limit used for limiting depth or degree of nodes that are added
+ * to the queue.
  */
-abstract class DistanceTraverser[T](graph: Graph, homeNodeIds: Seq[Int], onlyOnce: Boolean)
-    extends Traverser { self =>
+class Limit(limit: Option[Int]) {
```
szymonm: For some reason this is not possible.
What about:
```
object Walk {
  trait Limit {...}
  object NodeColor {...}
  trait NodeColoring {...}
  case class Limits {...}
}
```
?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,420 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+   */
+  def shouldBeEnqueued(color: NodeColor.Color): Boolean
+
+  /**
+   * Options to limit traverser walk.
+   */
+  lazy val limits: WalkLimits = WalkLimits()
+
+  lazy val depthLimit: Limit = new Limit(limits.maxDepth)
+  lazy val degreeLimit: Limit = new Limit(limits.maxNumEdgesThreshold.map(_ + 1))
```
szymonm: The API before was so that degreeLimit was exclusive and depth inclusive. Changed the API. Now both are inclusive as limit word suggests (as for me at least)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: got it, np.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/tourist/PrevNbrCounter.scala
pankajgupta: ok, makes sense.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: so, lets call it discoveryTime then as per the references?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,461 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+
+/**
+ * Trait to be mixed in to the DFS traverser to add finishing distance of a node.
+ */
+trait FinishingTimeTracker extends DepthFirstTraverser {
+  private var time = 0
+
+  private lazy val finishingTime = new IntInfoKeeper(true)
+
+  override def handleVisitedInQueue(next: Int): Unit = {
```
pankajgupta: this doesn't work I think. e.g., in the simple graph 1->2, handleVisitedInQueue is never called (nodes 1 and 2 are not visited again) so no finish times are ever recorded. 

Note that '1' is finished after '2' has been visited. Hence my previous comment: "What we should do in DFS is to not dequeue 'v' when we process it (i.e., in next()), but only in findNextNodeToVisit(). Then the time we first dequeue a node is its finish time. Does that seem right?"

btw, we should just put finishingTime inside DFS rather than a separate trait.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,461 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+    finishingTime.recordInfo(next, time)
+    super.handleVisitedInQueue(next)
+  }
+
+  override def visitNode(node: Node): Unit = {
+    time += 1
+    // adding again to capture finishing time of a node
+    queue.enqueueFirst(node.id)
```
pankajgupta: oh I missed that you are enqueuing again. ok but please still make calculation of finishingtime a part of the main DFS class instead of a separate trait. 

rather than wastefully deq+enq, can we just not do any operation at all for DFS while still deqing in BFS?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: I like the solution with separate trait very much. The caller code looks clear:
```
val trav = new DepthFirstTraverser(TestGraphs.g5, GraphDir.OutDir, Seq(10)) with FinishingTimeTracker
```
You don't need to keep the map with finishing time if you don't need to. The same with PathLengthTracker or DiscoveryTimeTracker (I think we can add it easily). And we can combine any of them using traits. 

The code of FinishingTimeTracker itself is also very clear and we don't mix all the things in the main DFS code.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: That may be a little wastefull, but:
1) We cannot deque in the findNextToDeque, because we need to have hasNext that does not dequeu. So we deque in the QueueBasedTraverser:next.
2) We can override QueueBasedTraverser:next in DepthFirstSearch or FinishingTimeTracker but wastefull adding when visiting looks more clear and concise for me.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: But that is still not time, but depth, maybe discoveryDepth?

szymonm:
@pankajgupta fixed small issues.
We still have to discuss 2 things. 
Traits?: I added trait for DiscoveryTime in DFS to convince you that traits here are not so bad. Look at the Spec - I think it looks very elegant now. And user can choose which Tracker(s) to use.
Naming: Time vs Depth: In both DFS and BFS.

pankajgupta:
Cool. I'm ok with traits. Will review in a few hours the rest.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: 'time' is fine. using conventional notation makes it clearer to a reader.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,487 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
 
-  init(-1)
+  def distance(id: Int) = visitDistanceTracker.infoOfNode(id)
+}
 
-  protected def storeForVisit(id: Int, info: Int) { stack.add(NodeDesc(id, info)) }
+/**
+ * Trait to be mixed in to the DFS traverser to add discovery time of a node.
+ */
+trait DiscoveryTimeTracker extends DepthFirstTraverser {
```
pankajgupta: shouldn't this (as well as FinishingTimeTracker) extend QueueBasedTraverser so that it can be used in BFS as well?

cassovary-core/src/test/scala/com/twitter/cassovary/graph/TraverserSpec.scala
```diff
@@ -177,42 +181,68 @@ class TraverserSpec extends Specification with Mockito {
...
+      List(10, 11) foreach { bfs.depth(_) mustEqual Some(1) }
+      List(12) foreach { bfs.depth(_) mustEqual Some(2) }
+      List(13, 16) foreach { bfs.depth(_) mustEqual None }
+    }
   }
 
   "DepthFirstTraverser" should {
-    val graph = TestGraphs.g6
+    "yield all nodes in DFS order with expected finishing times" in {
+      val trav = new DepthFirstTraverser(TestGraphs.g5, GraphDir.OutDir, Seq(10))
```
pankajgupta: looks good. please also add test with these traits in BFS.

szymonm:
I added DiscoveryTime to BFS and 1 test for it. However, it's not very usefull because discovery time is the same as order of nodes in the BFS.

FinishingTime cannot be so easily used in the BFS (we need similar checking and deleting visited nodes in the queue as in DFS), so I'm wondering if this is really worth it?

pankajgupta:
Agreed finish time is not relevant for BFS.

pankajgupta:
One last ting -- isn't 'time' in DiscoveryTime equal to numEnqueuedEver? If so, we can get rid of it.

szymonm:
This is what I meant. In BFS, time in discoveryTime of last node is equal to numEnquedEver and every node time is the index of the node in the traverser. Or, equivalently time == numEnquedEver. Thus, mixing the DiscoveryTimeTracker trait to BFS makes little sense, I think. Nevertheless it is possible and in this setting DiscoveryTimeTracker trait gives consecutive numbers to visited nodes. See TraverserSpec:183

szymonm:
If everything is fine, let me squash the commits into one.

pankajgupta:
what I meant was that if time equals numEnquedEver, why have a separate variable at all? DiscoveryTimeTrackerTrait should just record numEnqueuedever.

szymonm:
Hi in BFS time is indeed equal to numEnquedEver. Things are more problematic in the DFS, where a node can be queue more than once - this does not work.

szymonm:
I squashed all into one commit.

szymonm:
This is how it looks without dequeing.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,487 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
-        // try the next child
-        firstNotYetVisitedNode(nd, off + 1)
-      } else {
-        storeForVisit(nd.id, off)
-        Some(NodeDesc(childId, distance(nd.id).get + 1))
-      }
+  abstract override protected def enqueue(nodes: Seq[Int], from: Option[Int]): Unit = {
+    nodes.foreach {
+      node =>
+        if (discoveryTime(node).isEmpty) {
```
pankajgupta: no need to check because discoveryTime has onlyOnce set to true

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,487 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
-  def next = {
-    var NodeDesc(id, off) = stack.remove(stack.size - 1)
-    assert(off == -1)
-    val topnd = getExistingNodeById(graph, id)
-    val nextUnvisitedNode = firstNotYetVisitedNode(topnd, off + 1)
-    nextUnvisitedNode match {
-      case Some(NodeDesc(nodeId, dist)) => visitPotentialNode(nodeId, -1, dist)
-      case None => // do nothing
+  override def handleVisitedInQueue(node: Int): Unit = {
+    if (finishingTime(node).isEmpty) {
```
pankajgupta: no need to check

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: do need to check whether to increase time. I thought it's more clear to check it before recording.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: as before.

We can also set onlyOnce to false.

szymonm:
Needed to rebase, becauase of #104

szymonm:
Hey @pankajgupta, can you look at this?

pankajgupta:
will do in a day

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,492 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
-  def distance(id: Int) = distanceTracker.infoOfNode(id)
+  def hasNext = findNextNodeToVisit().isDefined
 }
 
 /**
- * Traverses in breadth-first order. This implies that first all the neighbors of
- * {@code homeNodeIds} are visited, then their neighbors, etc.
+ * A traverser that keeps track of first depth of visiting a given node.
+ */
+trait DepthTracker extends QueueBasedTraverser {
```
pankajgupta: May be a small point but should all trackers be of type mixin traits? i.e., trait DepthTracker { this: QueueBasedTraverser => ... } rather than extending QueueBasedTraverser.
http://baddotrobot.com/blog/2014/09/22/scala-mixins/

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,492 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
+/**
+ * Trait to be mixed in to the BFS/DFS traverser to add discovery and finishing
+ * times of a node.
+ *
+ * Discovery time of a node is time when the node is added to the queue for the first time.
+ *
+ * Finishing time of a node is time just after all `dir` neighbors of the node became finished
+ * (or when it is visited if it has no `dir` neighbors that should be processed later).
+ */
+trait DiscoveryAndDepthTimeTracker extends DepthFirstTraverser {
```
pankajgupta: s/DiscoveryAndDepthTime/DiscoveryAndFinishTime/

again a self type trait might be better

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -119,166 +124,492 @@ class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
...
-      if (onlyOnce && seenBefore(childId)) {
-        // try the next child
-        firstNotYetVisitedNode(nd, off + 1)
-      } else {
-        storeForVisit(nd.id, off)
-        Some(NodeDesc(childId, distance(nd.id).get + 1))
-      }
+  override val shouldBeDequeuedBeforeProcessing = false
+
+  abstract override protected def enqueue(nodes: Seq[Int], from: Option[Int]): Unit = {
```
pankajgupta: in all trackers we seem to be doing something before enqueue(). Consider having an abstract preEnqueueHook() in QueueBasedTraverser which is implemented by the tracker traits.

pankajgupta:
pending the above changes, rest LGTM

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: I'm not quite sure. I'm using here stackable trait pattern (http://www.artima.com/scalazine/articles/stackable_trait_pattern.html, http://blog.ryannedolan.info/2014/03/scala-pattern-stackable-traits.html). On both this sites people use extending. Type constraints are often used for Cake Pattern (http://stackoverflow.com/questions/1990948/what-is-the-difference-between-self-types-and-trait-subclasses) that is often connected to Dependency Injection and seems to be closer to composition than inheritence. Is this convincing?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: the same as before I think

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: The preEnqueueHook should then be able to change the nodes to enqueue, it should have the from node information (for depth for example). We obtain:
```
preEnqueueHook(nodes: Seq[Int], from: Option[Int]): (Seq[Int], Option[Int]) = {...}
```
Enqueu is currently called in QueueBasedTraverser:369 in processNode. That would be
```
enqueue(preEnqueueHoook(chooseNodesToEnqueue(node), Some(node.id)))
```
Looking at the stackable trait pattern articles - they usually use abstract override methods like a pre...Hook without separate method.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: ok, this is fine.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: I think the current thing is okay.

PR #94, opened 2015-01-04
SmallBoundedPriorityQueueSpec migration to scalatest

bartekkalinka:
Another test migration


szymonm:
LGTM.
@bartekkalinka Could you do a rebase and squash this 3 commits into one?
@pankajgupta I think it's ready to merge after rebase.

bartekkalinka:
Done

cassovary-core/src/test/scala/com/twitter/cassovary/util/SmallBoundedPriorityQueueSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.util
+
+import org.scalatest.WordSpec
+import org.scalatest.matchers.ShouldMatchers
+
+class SmallBoundedPriorityQueueSpec extends WordSpec with ShouldMatchers {
+  "SmallBoundedPriorityQueue of 3 ints" when {
+    val q = new SmallBoundedPriorityQueue[Int](3)
```
pankajgupta: the problem with doing this is that these tests can not be done in parallel because of shared mutable state held by the instance val. see http://doc.scalatest.org/2.2.1/#org.scalatest.WordSpec and search for "Note that the only way before and after code can communicate with test code".

One option is to change this to a 'def fixture = new ...' and inserting 'val q = fixture' at the beginning of each "in" block

cassovary-core/src/test/scala/com/twitter/cassovary/util/SmallBoundedPriorityQueueSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+        q.size shouldEqual 0
+        q += 1
+        q += 10
+        q.top(2).toList shouldEqual List(10, 1)
+      }
+    }
+  }
+
+  "SmallBoundedPriorityQueue of 5 strings" when {
+    val q = new SmallBoundedPriorityQueue[String](5)
```
pankajgupta: same as above

bartekkalinka:
I don't think the "in" blocks were meant to be executed in parallel.  For example, "not full" and "full" blocks were one sequential block before the modification, and the latter depends on the results from the former.  So, should I:
- somehow mark these "in" blocks for sequential exection (is it possible and how)?
- redesign these tests so each "in" block can be executed in parallel (involves more changes than modifying val q)

I had no idea I had to consider parallel execution here.  Whole "when" blocks could be run in parallel  now though (I think).

szymonm:
I used fixtures here:
https://github.com/twitter/cassovary/blob/master/cassovary-core/src/test/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReaderSpec.scala

bartekkalinka:
So "in" blocks should be independent of each other?  I'll rewrite whole thing.

pankajgupta:
yes, that is best practice.

bartekkalinka:
Done, let me know what you think.

cassovary-core/src/test/scala/com/twitter/cassovary/util/SmallBoundedPriorityQueueSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+        q.top(3).toList shouldEqual List(30, 10, 4)
+        q += 100
+        q += 1
+        q += 10
+        q.top(3).toList shouldEqual List(100, 30, 10)
+      }
+
+      "return correct number of elements" in {
+        val q = fullQueue[Int](List(4, 3, 30))
+        q.top(1).toList shouldEqual List(30)
```
pankajgupta: did you mean to test something else? because the test description says it is testing "number of elements"

cassovary-core/src/test/scala/com/twitter/cassovary/util/SmallBoundedPriorityQueueSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+    content.foreach(q += _)
+    q
+  }
+
+  "SmallBoundedPriorityQueue of 3 ints" when {
+
+    "empty" should {
+      "return empty list" in {
+        val q = emptyQueue[Int](3)
+        q.top(0).toList shouldEqual List()
```
pankajgupta: I know the original test converted everything to List() and then checked matching. But can you please check whether that can be avoided here? i.e., whether 
q.top(0) shouldEqual Seq() will work
etc. throughout this test.

bartekkalinka:
Seq works fine.  Corrected test description.

PR #95, opened 2015-01-04
Migrate DirectedGraphSpec from spec to ScalaTest

adamkozuch:
My first migrated test

### DIFF
--- a/cassovary-core/src/test/scala/com/twitter/cassovary/graph/DirectedGraphSpec.scala
+++ b/cassovary-core/src/test/scala/com/twitter/cassovary/graph/DirectedGraphSpec.scala
@@ -24,50 +24,47 @@ class DirectedGraphSpec extends WordSpec with ShouldMatchers {
   val twoNodeGraph = TestGraphs.g2_mutual
   val sixNodeGraph = TestGraphs.g6
 
-  "two node graph with each following the other" should {
-    "nodeCount and edgeCount should be correct" in {
-      twoNodeGraph.nodeCount shouldEqual 2
-      twoNodeGraph.edgeCount shouldEqual 2
+  "DirectedGraph" when {
+    "two node with each following the other" should {
+      "nodeCount and edgeCount be correct" in {
+        twoNodeGraph.nodeCount shouldEqual 2
+        twoNodeGraph.edgeCount shouldEqual 2
+      }
+
+      "maxNodeId be correct" in {
+        twoNodeGraph.maxNodeId shouldEqual 2
+      }
+
+      "elements return all the nodes" in {
+        twoNodeGraph.iterator.toList.map {_.id }.toSet shouldEqual Set(1, 2)
+      }
+
+      "getNodeById work in both positive and negative cases" in {
+        twoNodeGraph.getNodeById(1) shouldEqual Some(TestNode(1, List(2), List(2)))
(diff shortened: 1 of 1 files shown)


cassovary-core/src/test/scala/com/twitter/cassovary/graph/DirectedGraphSpec.scala
```diff
@@ -13,58 +13,60 @@
...
     "maxNodeId should be correct" in {
-      graph.maxNodeId mustEqual 2
+      twoNodeGraph.maxNodeId shouldEqual 2
     }
 
     "elements should return all the nodes" in {
-      graph.iterator.toList.map { _.id }.sortBy (+_) mustEqual List(1, 2)
+      twoNodeGraph.iterator.toList.map {
+        _.id
+      }.sortBy(+_) shouldEqual List(1, 2)
```
szymonm: What's the point in sortBy(+_). Isn't sorted better? Or, even better, toSet shouldEqual Set(1,2)

cassovary-core/src/test/scala/com/twitter/cassovary/graph/DirectedGraphSpec.scala
```diff
@@ -13,58 +13,60 @@
...
 
     "maxNodeId should be correct" in {
-      graph.maxNodeId mustEqual 2
+      twoNodeGraph.maxNodeId shouldEqual 2
     }
 
     "elements should return all the nodes" in {
-      graph.iterator.toList.map { _.id }.sortBy (+_) mustEqual List(1, 2)
+      twoNodeGraph.iterator.toList.map {
+        _.id
```
szymonm: I don't like new lines here. I would write it in one line as there is no serious logic in the _.id. And probably only in ( ).

cassovary-core/src/test/scala/com/twitter/cassovary/graph/DirectedGraphSpec.scala
```diff
@@ -13,58 +13,60 @@
...
 
     "maxNodeId should be correct" in {
-      graph.maxNodeId mustEqual 15
+      sixNodeGraph.maxNodeId shouldEqual 15
     }
 
     "elements should return all the nodes" in {
-      graph.iterator.toList.map { _.id }.sortBy(+_) mustEqual List(10, 11, 12, 13, 14, 15)
+      sixNodeGraph.iterator.toList.map {
+        _.id
```
szymonm: as before

szymonm:
Hi @adamkozuch, I'm glad to see your change.

This test, however, was a bit broken from the beginning. The specification written as strings does not fit well to the WordSpec schema. This results in output like:
```
[info] + two node graph with each following the other should
[info]   + nodeCount and edgeCount should be correct
[info]   + maxNodeId should be correct
[info]   + elements should return all the nodes
[info]   + getNodeById should work in both positive and negative cases
```
that is not fully language correct. I think it's possible to fix it using WordSpec 'when' special word. You can check the documentation here: http://doc.scalatest.org/1.9.1/#org.scalatest.WordSpec

Could you try to fix this issues?

adamkozuch:
I made changes. Is it correct now?

szymonm:
Thanks, LGTM @pankajgupta

PR #96, opened 2015-01-05
Migrate two more specs to scalatest.

pankajgupta:


cassovary-core/src/test/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNodeSpec.scala
```diff
@@ -13,33 +13,37 @@
...
-import org.specs.Specification
+import org.scalatest.WordSpec
+import org.scalatest.matchers.ShouldMatchers
+
+class SynchronizedDynamicNodeSpec extends WordSpec with ShouldMatchers {
+
+  def fixture(id: Int) = new SynchronizedDynamicNode(id)
 
-class SynchronizedDynamicNodeSpec extends Specification {
   "Two SynchronizedDynamicNodes equal iff their nodeIds are equal" in {
```
szymonm:
```
"Two SynchronizedDynamicNodes" should {
    "equal iff their nodeIds are equal" in {
   ...
   }
   "add/delete functions perform correctly" in {
   }
```
would be better I think

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/PageRankSpec.scala
```diff
@@ -13,53 +13,53 @@
...
 
     "Return the correct values with 1 iteration" in {
       val params = PageRankParams(0.9, Some(1))
       val pr = PageRank(graphG6, params)
-      pr must MapAlmostEquals(Map(10 -> (.1/6 + .9/12), 11 -> (.1/6 + .9*(1.0/18+1.0/12)),
+      pr should almostEqualMap(Map(10 -> (.1/6 + .9/12), 11 -> (.1/6 + .9*(1.0/18+1.0/12)),
         12 -> (.1/6 + .9*(1.0/6+1.0/18)), 13 -> (.1/6 + .1/2), 14 -> (.1/6 + .9/3), 15 -> 1.0/6))
     }
 
     "At 2 iterations PageRank still sums to 1" in {
```
szymonm: I would write "At 2 iterations still sum to 1" (second part of PageRank should...)

cassovary-core/src/test/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNodeSpec.scala
pankajgupta: yes, changed.

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/PageRankSpec.scala
pankajgupta: yes, good point. changed.

PR #103, opened 2015-01-10
Migrate DirectedGraphUtils and DirectedPathCollection to ScalaTest

adamkozuch:


cassovary-core/src/test/scala/com/twitter/cassovary/graph/DirectedGraphUtilsSpec.scala
```diff
@@ -13,65 +13,72 @@
...
-    }
-  }
+  "DirectedGraphUtils" when {
 
-  "mutual edges on six node graph with only one dir stored" should {
-    "be correct with only out dir" in {
-      val directedGraphUtils = new DirectedGraphUtils(TestGraphs.g6_onlyout)
-      directedGraphUtils.getNumMutualEdges mustEqual 0L
+    "two node with each following the other" should {
+      val (graph, directedGraphUtils) = utils(TestGraphs.g2_mutual)
```
szymonm: this line can be in the next scope

PR #108, opened 2015-01-12
Migrated ArrayBasedDirectedGraph and SharedArrayBasedDirectedGraph.

pankajgupta:
Did some refactroing in the process.

Hey @szymonm can you do a quick review?
Also, logging a followup issue -- the file GraphBehaviours uses asserts but it would be better to use ShouldMatchers. This way the tests fail gracefully.


pankajgupta:
just renamed.

PR #112, opened 2015-01-15
GraphUtils migration to ScalaTest

szymonm:
Did an easy migration to ScalaTest. However, the test is incomplete mainly because the point of GraphUtils is not very clear to me. In many cases GraphUtils is only a wrapper for concrete algorithms.


pankajgupta:
LGTM. 

(I understand what you are saying. The idea is to provide the glue to make the algorithm functions a little more friendly to use. That may or may not be actually happening. Indeed is no point in having a dumb wrapper. Lets treat that issue separately. Happy to take your suggestions on removing functions one by one in graphutils).

PR #113, opened 2015-01-15
Changed asserts to ShouldMatchers in GraphBehaviours

AnishShah:


cassovary-core/src/test/scala/com/twitter/cassovary/graph/GraphBehaviours.scala
```diff
@@ -15,31 +15,34 @@ package com.twitter.cassovary.graph
...
     "have number of edges less than in full graph" in {
-      assert(graph.edgeCount <= (numNodes * (numNodes - 1).toLong))
+      graph.edgeCount should be <= (numNodes * (numNodes - 1).toLong)
     }
   }
 
   def completeGraph(graph: => DirectedGraph, numNodes: Int) {
     "have correct number of nodes" in {
-      assert(graph.nodeCount === numNodes, "wrong number of nodes")
+      withClue("wrong number of nodes") {
```
pankajgupta: no need for this extra statement. just the shouldEqual statement is enough.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/GraphBehaviours.scala
```diff
@@ -51,16 +54,26 @@ trait GraphBehaviours {
       graph foreach { node =>
         (0 to 1) foreach { dir =>
           val neighbors = if (dir == 0) node.outboundNodes() else node.inboundNodes()
-          assert(neighbors.size === neighbors.toSet.size, "duplicates in neighbors array detected")
+          withClue("duplicates in neighbors array detected") {
```
pankajgupta: withClue is okay in this block because there are several conditions tested.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/GraphBehaviours.scala
AnishShah: Should I remove it everywhere ?

pankajgupta:
the use of withClue is okay in check(). Would be good for @szymonm to take a look to see where the use of withClue seems superfluous.

LGTM otherwise. @szymonm : feel free to merge once you are satisfied.

AnishShah:
Cool :+1:  Except the first `withClue` (now removed), other `withClue` gives detailed error if there's a failure since there are many conditions. So, it should be OK. What do you think ? @szymonm

szymonm:
LGTM.
@pankajgupta I don't have rights to merge.

PR #114, opened 2015-01-15
Bumped fastutil to 6.6.0. #93

### LINKED ISSUE #93: bump fastutils to the newest version
We use fastutil 6.4.4 while 6.5.16 is the current.

szymonm:
This resolves #93.
Improved DirectedPathComparator after fastutils map behaviour change

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPath.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPath.scala
@@ -19,7 +19,7 @@ import java.util.Arrays
 /**
  * Represents a directed path of nodes in a graph. No loop detection is done.
  */
-case class DirectedPath(val nodes: Array[Int]) {
+case class DirectedPath(nodes: Array[Int]) {
 
   /**
    * @return the number of nodes on this path
@@ -43,6 +43,10 @@ case class DirectedPath(val nodes: Array[Int]) {
 
   override def hashCode(): Int = {
     Arrays.hashCode(nodes)
+  }
+
+  override def toString(): String = {
+    "DirectedPath(" + nodes.mkString(",") + ")"
   }
 }
 
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
@@ -42,7 +42,7 @@ class DirectedPathCollection {
    */
   def appendToCurrentPath(node: Int) {
     currPath.append(node)
-    pathCountsPerIdWithDefault(node).add(currPath.snapshot, 1)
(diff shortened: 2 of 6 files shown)


szymonm:
@pankajgupta, can you still look at it? I needed to improve directed path comparator. Would be good if you looked at it.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
```diff
@@ -148,10 +147,18 @@ class PathCounterComparator(pathCountsPerId: Int2ObjectOpenHashMap[Object2IntOpe
...
+    if (dp1Count != dp2Count) {
+      if (descending) {
+        dp2Count - dp1Count
+      } else {
+        dp1Count - dp2Count
+      }
     } else {
-      dp1Count - dp2Count
+      if (descending) {
+        dp2.length - dp1.length
```
pankajgupta: might be counter-intuitive, but shorter paths are generally better, so I would reverse this -- i.e., put dp1.length - dp2.length here and put a brief comment here as well

cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
```diff
@@ -54,11 +54,10 @@ class DirectedPathCollection {
 
   /**
    * @return an map of top DirectedPaths with occurrence
-   *         ending at {@code node}, sorted decreasing
+   *         ending at `node`, sorted decreasing
```
pankajgupta: comment is still not clear. How about
@return a map of {path -> count} for paths ending at node, sorted in decreasing order of count. Paths with the same count are sorted shortest path first.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPath.scala
```diff
@@ -44,6 +44,10 @@ case class DirectedPath(val nodes: Array[Int]) {
   override def hashCode(): Int = {
     Arrays.hashCode(nodes)
   }
+
+  override def toString(): String = {
```
pankajgupta: I would remove explicit type annotation here since it is so clear from the body

project/Build.scala
```diff
@@ -6,6 +6,8 @@ import SonatypeKeys._
 
 object Cassovary extends Build {
 
+  val fastUtilsVersion = "6.6.0"
```
pankajgupta: how about refactor the whole dependency Seq("it.unimi.dsi" % "fastutil" % "6.6.0") ? And while at it, refactor the other common parts (settings, name, etc.) as well of all the non cassovary-core projects in this file? ok with me if you want to do that as the same or different PR

pankajgupta:
I have reverted the merge for now to allow you to modify as per above comments.

PR #116, opened 2015-01-15
Minor improvements to SynchronizedDynamicGraph

plofgren:
To get my feet wet contributing to Cassovary I'd like to make two improvements to SynchronizedDynamicGraph:
1. A minor fix in toString (the in and out neighbors had been mislabled)
2. Adding support for node id 0
The justification for the second one is that in another project I wrote a small example graph with node ids 0, 1, and 2, and SynchronizedDynamicGraph silently ignored the edges to node 0, causing strange results that took time to debug (especially given the unexpected toString behavior).

Thanks!

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNode.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNode.scala
@@ -22,7 +22,9 @@ import scala.collection.mutable.ArrayBuffer
 class SynchronizedDynamicNode(val id: Int) extends DynamicNode {
   protected val inEdges = new ArrayBuffer[Int]
   protected val outEdges = new ArrayBuffer[Int]
-  val DELETED_MARKER = 0
+  val DELETED_MARKER = Integer.MIN_VALUE
+  if (id == DELETED_MARKER)
+    throw new IllegalArgumentException("ID " + DELETED_MARKER + " is not allowed.")
 
   /**
    * Override functions of base class.
@@ -85,6 +87,6 @@ class SynchronizedDynamicNode(val id: Int) extends DynamicNode {
   override def toString = {
     val inNodes = synchronized { inboundNodes.toList }
     val outNodes = synchronized { outboundNodes.toList }
-    "SynchronizedDynamicNode(id=%d, out=%s, in=%s)".format(id, inNodes, outNodes)
+    "SynchronizedDynamicNode(id=%d, out=%s, in=%s)".format(id, outNodes, inNodes)
   }
 }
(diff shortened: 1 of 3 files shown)


cassovary-core/src/test/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNodeSpec.scala
```diff
@@ -46,6 +46,13 @@ class SynchronizedDynamicNodeSpec extends WordSpec with ShouldMatchers {
       node.removeOutBoundNode(4)
       node.outboundNodes.toList shouldEqual List(2)
     }
+
+    "throw an exception if id is DELETED_MARKER" in {
+      val DELETED_MARKER = fixture(1).DELETED_MARKER
```
pankajgupta: reference the object here not fixture(1)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNode.scala
```diff
@@ -22,7 +22,9 @@ import scala.collection.mutable.ArrayBuffer
 class SynchronizedDynamicNode(val id: Int) extends DynamicNode {
   protected val inEdges = new ArrayBuffer[Int]
   protected val outEdges = new ArrayBuffer[Int]
-  val DELETED_MARKER = 0
+  val DELETED_MARKER = Integer.MIN_VALUE
```
pankajgupta: I know this was already there, but it will be better to put this in a companion object to save space.

i.e., object SynchronizedDynamicNode {
val DELETED_MARKER = ...
}

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNode.scala
```diff
@@ -22,7 +22,9 @@ import scala.collection.mutable.ArrayBuffer
...
+  if (id == DELETED_MARKER)
```
pankajgupta: use require() instead

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNode.scala
plofgren: Good, I was wondering why it was in the class and not in the object.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNode.scala
plofgren: Nice, I didn't know about require.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/node/SynchronizedDynamicNodeSpec.scala
plofgren: Right, moving DELETED_MARKER to companion object makes this cleaner.

Issue #64, opened 2014-04-28
move *Spec tests in src/test (one by one) from 'specs' to scalatest


pankajgupta:
'org.specs' has been deprecated and the tests should move to using org.scalatest. An example of the latter is in com.twitter.cassovary.graph.TestGraphSpec.

pankajgupta:
Since someone asked, here is the online documentation for ScalaTest -- we are using WordSpec and ShouldMatchers here.
http://doc.scalatest.org/2.2.1/#org.scalatest.WordSpec
http://www.scalatest.org/user_guide/using_matchers

Can also check out recent pull requests https://github.com/twitter/cassovary/pull/94 or https://github.com/twitter/cassovary/pull/96 to see the difference between Specs and Scalatest

Also, please fill the sheet mentioned by @szymonm above as multiple people are working on this issue.

pankajgupta:
Fixed and removed specs in PR #117

PR #117, opened 2015-01-16
Removed specs post migration of tests to scalatest.

pankajgupta:
After merging this, issue https://github.com/twitter/cassovary/issues/64 will be resolved.


pankajgupta:
please take a look @szymonm

PR #123, opened 2015-01-22
Move to Scala 2.11

caniszczyk:
Signed-off-by: Chris Aniszczyk <zx@twitter.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: scala
 sudo: false
 scala:
-  - 2.10.3
+  - 2.11.4
--- a/cassovary-core/src/test/scala/com/twitter/cassovary/util/MapNodeNumbererSpec.scala
+++ b/cassovary-core/src/test/scala/com/twitter/cassovary/util/MapNodeNumbererSpec.scala
@@ -17,10 +17,10 @@ package com.twitter.cassovary.util
 import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.WordSpec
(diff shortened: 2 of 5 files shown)


travisbrown:
Hey @caniszczyk, it'd probably be best to set this up to continue cross-publish 2.10 as well, and we might as well bump to 2.11.5 while we're at it. I can file a PR in a second with the changes.

caniszczyk:
@travisbrown I'm trying to avoid cross publishing in this case, I don't think we need too IMHO, 2.10 is already pretty old :)

szymonm:
@caniszczyk You also need to bump Scalatest to newest version.

travisbrown:
And `finagle-stats`, since we don't have 2.11 builds until 6.24.0.

caniszczyk:
@travisbrown done

pankajgupta:
hey @caniszczyk , I think I agree with @travisbrown that it would be good to continue cross publishing scala 2.10 .

pankajgupta:
We can deprecate scala 2.10 in a few months...

pankajgupta:
@caniszczyk travis ci build is failing for this PR.

caniszczyk:
@pankajgupta yep looks like some changes in the matchers package since scalatest 2.10, looking to fix it

caniszczyk:
The specs are fixed now to run with 2.11, not sure why Travis CI doesn't want to build for me

pankajgupta:
Thanks for fixing! Thoughts on continuing to crosspublish for scala 2.10 as well?

caniszczyk:
@pankajgupta I created #126 for that, seems fairly easy to do

Issue #126, opened 2015-01-23
Cross publish Scala 2.10/2.11 releases


caniszczyk:

caniszczyk:
I'll wait for @pankajgupta to do the honor of publishing a new release

pankajgupta:
Thanks @caniszczyk . Noticed that there are a fair amount of 'evicted' warnings in sbt. Should we try to fix them? e.g., this one is quite frequent:
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn] 	* com.google.guava:guava:13.0.1 -> 16.0.1 (caller: com.twitter.common:collections:0.0.82, com.twitter.common:quantity:0.0.79, com.twitter:util-collection_2.11:6.23.0, com.twitter.common:util-system-mocks:0.0.81, com.twitter.common:stats-provider:0.0.66, com.twitter.common:jdk-logging:0.0.54, com.twitter.common:base:0.0.95, com.twitter.common:stats-util:0.0.49, com.twitter:finagle-http_2.11:6.24.0, com.twitter.common:stats:0.0.98, com.twitter.common:util:0.0.106, com.twitter.common:util-executor-service-shutdown:0.0.55, com.github.spullara.mustache.java:compiler:0.8.12.1, com.twitter:twitter-server_2.11:1.9.0, com.twitter.common:util-sampler:0.0.61, com.twitter:finagle-stats_2.11:6.24.0, com.twitter.common:metrics:0.0.29, com.twitter:cassovary-server_2.11:4.0.0, com.twitter:cassovary-core_2.11:4.0.0, com.twitter.common:application-action:0.0.75)

pankajgupta:
Released, thanks @caniszczyk and @travisbrown .

PR #127, opened 2015-01-23
Update Scala 2.11 version and enable 2.10 cross-publishing

travisbrown:
This does a few little clean-up chores: it adds support for cross-publishing for 2.10 (see #126), updates the Travis CI config to test both Scala versions (and multiple JDKs), and adds explicit SBT version configuration for people who aren't using the provided `sbt` script.

Note that I'm not using the provided `sbt` script in the Travis build because of [this issue](https://github.com/travis-ci/travis-ci/issues/3120). When they get that fixed it's probably worth reconsidering which we want to use there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,14 @@
 language: scala
 sudo: false
+
 scala:
-  - 2.11.4
+  - 2.10.4
+  - 2.11.5
+
+jdk:
+  - openjdk7
+  - oraclejdk7
+  - oraclejdk8
+
+script:
+  - sbt ++$TRAVIS_SCALA_VERSION test
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -36,9 +36,6 @@ object Cassovary extends Build {
     scalacOptions ++= Seq("-encoding", "utf8"),
     scalacOptions += "-deprecation",
 
-    javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
-    javacOptions in doc := Seq("-source", "1.6"),
-
     fork in run := true,
     javaOptions in run ++= Seq("-server"),
     outputStrategy := Some(StdoutOutput),
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -17,7 +17,8 @@ object Cassovary extends Build {
   val sharedSettings = Seq(
     version := "4.0.0",
     organization := "com.twitter",
-    scalaVersion := "2.11.4",
+    scalaVersion := "2.11.5",
+    crossScalaVersions := Seq("2.10.4", "2.11.5"),
     retrieveManaged := true,
     libraryDependencies ++= Seq(
       "com.google.guava" % "guava" % "16.0.1",
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version=0.13.7


caniszczyk:
Should we just toss the sbt script? I find it silly that we have it, we should assume sbt exists on the users computer. 

Also, is our project/Build.scala the right way to do things? I've noticed a "build.sbt" file at the root of projects these days and wondering if that's the new way of structuring SBT builds.

.travis.yml
```diff
@@ -1,4 +1,14 @@
...
 sudo: false
+
 scala:
-  - 2.11.4
+  - 2.10.4
+  - 2.11.5
+
+jdk:
+  - openjdk7
+  - oraclejdk7
```
caniszczyk: I'd drop oraclejdk7, we should only need one v7 JDK IMHO, otherwise just a waste of build resources in the majority of cases

travisbrown:
For projects like Finagle the custom SBT launcher script provides some additional tuning (as well as the convenience of not requiring the user to have SBT installed, which I agree isn't enough to justify the script). I'm not familiar enough with this project to know whether the tuning provides any value here.

And right, with recent versions of SBT the `build.sbt` format is fixed for multi-project builds, and it's a lot cleaner and more concise than `Build.scala`. I've been migrating projects as I find time—it's not too hard to do, but it's non-trivial enough that I didn't want to do it in this PR.

.travis.yml
travisbrown: I'm fine with this, but I'm pretty sure I've seen issues come up only with that JDK on other projects, and since it's currently probably the most widely used of the three, I thought it would be worth testing.

As a side note, I'd like to standardize these points of Travis CI configuration across our OSS Scala projects, and right now we're testing all three for most projects. If we want to drop Oracle's JDK 7 we should probably do it across the board.

szymonm:
Do we stil need:
```
    javacOptions ++= Seq("-source", "1.6", "-target", "1.6"),
    javacOptions in doc := Seq("-source", "1.6"),
```
then?

pankajgupta:
1. We had a build.sbt but migrated to Build.scala when we moved to multiple sub-projects. We can move back if you guys think it's a good idea. Any examples of twitter oss projects using the new Build.sbt?
2. We shouldn't need the new javac options as pointed by @szymonm 
3. custom sbt launcher: I have seen problems before with having a low heap size so at least the -Xms -Xmx options are useful.

.travis.yml
pankajgupta: If oraclejdk7 is the most widely used right now, we should probably keep it....

travisbrown:
@pankajgupta Yeah, there were lots of things that couldn't be done in `build.sbt` until recently, but now that's the blessed format, and in my experience it works really well (on 0.13.7). See these Finagle organization projects, for example: [finch](https://github.com/finagle/finch/blob/master/build.sbt), [webbing](https://github.com/finagle/webbing/blob/master/build.sbt), and [finagle-serial](https://github.com/finagle/finagle-serial/blob/master/build.sbt).

I'll remove the `javac` options in this PR once I'm back at my desk.

About the custom launcher—you can always set the heap size when you start SBT (or all the time on your machine) without the script, but if this is generally desirable I'd vote to leave it.

PR #130, opened 2015-01-24
Remove warnings post move to scala 2.11

AnishShah:


AnishShah:
Working on `SynchronizedBuffer` warning. Should I use `java.util.concurrent.ConcurrentLinkedQueue` as the Deprecation warning says ?

szymonm:
I think so.

AnishShah:
This removes all the warnings, except 

```
[warn] There may be incompatibilities among your library dependencies.
[warn] Here are some of the libraries that were evicted:
[warn] 	* com.google.guava:guava:13.0.1 -> 16.0.1
[warn] Run 'evicted' to see detailed eviction warnings
```

pankajgupta:
Filed separate issue https://github.com/twitter/cassovary/issues/132 for the library eviction warnings. Perhaps @caniszczyk or @travisbrown have seen that elsewhere as well?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
```diff
@@ -15,15 +15,16 @@ package com.twitter.cassovary.graph
...
+import scala.collection.JavaConverters._
 import scala.util.Random
-import scala.collection.mutable.ArrayBuffer
 
 /**
  * A simple implementation of a DirectedGraph
  */
 case class TestGraph(nodes: Node*) extends DirectedGraph {
-  val nodeTable = new mutable.HashMap[Int, Node]
+  val nodeTable = new HashMap[Int, Node]
```
pankajgupta: it is considered best practice to explicitly specify mutable.HashMap and import only scala.collection.mutable (see http://docs.scala-lang.org/overviews/collections/overview.html starting from "A useful convention is...")

cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
```diff
@@ -15,15 +15,16 @@ package com.twitter.cassovary.graph
 
 import com.twitter.cassovary.graph.StoredGraphDir._
 import com.twitter.cassovary.util.{Sampling, BinomialDistribution}
-import scala.collection.mutable
+import java.util.concurrent.ConcurrentLinkedQueue
+import scala.collection.mutable.HashMap
```
pankajgupta: delete (see next comment)

pankajgupta:
LGTM otherwise

PR #133, opened 2015-01-25
Resolves #110.

### LINKED ISSUE #110: Create a graph which keeps neighboring nodes sorted by id
In some applications one needs fast intersections of adjacency lists and fast search of whether a node v is a neighbor of a node u. In such cases, a graph that keeps the adjacency list in sorted order would be great.

This should work for both in and out directions in both ArrayBasedDirectedGraph and SharedArrayBasedDirectedGraph.

szymonm:
Allows nodes that store neighbors sorted in ArrayBasedGraph

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
@@ -293,10 +293,9 @@ object ArrayBasedDirectedGraph {
           val futures = (nodesOutEdges.iterator ++ Iterator(nodesWithNoOutEdges)).map {
             (nodes: Seq[Node]) => futurePool {
               nodes foreach { node =>
-                val fillingInEdgesNode = node.asInstanceOf[FillingInEdgesBiDirectionalNode]
                 val edgeSize = inEdgesSizes(node.id).intValue()
                 if (edgeSize > 0) {
-                  fillingInEdgesNode.createInEdges(edgeSize)
+                  node.asInstanceOf[FillingInEdgesBiDirectionalNode].createInEdges(edgeSize)
                   // reset inEdgesSizes, and use it as index pointer of
                   // the current insertion place when adding in edges
                   inEdgesSizes(node.id).set(0)
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
@@ -13,15 +13,17 @@
  */
 package com.twitter.cassovary.graph
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import com.google.common.annotations.VisibleForTesting
 import com.twitter.cassovary.graph.StoredGraphDir._
 import com.twitter.cassovary.graph.node._
-import com.twitter.cassovary.util.{SortedArrayOps, BoundedFuturePool}
+import com.twitter.cassovary.util.BoundedFuturePool
 import com.twitter.finagle.stats.DefaultStatsReceiver
 import com.twitter.logging.Logger
 import com.twitter.util.Future.when
-import com.twitter.util.{Await, FuturePool, Future}
-import java.util.concurrent.atomic.AtomicInteger
+import com.twitter.util.{Await, Future, FuturePool}
+
 import scala.collection.mutable
 
 /**
@@ -67,7 +69,7 @@ object NeighborsSortingStrategy extends Enumeration {
 }
 
 object ArrayBasedDirectedGraph {
-  import NeighborsSortingStrategy._
+  import com.twitter.cassovary.graph.NeighborsSortingStrategy._
 
   def apply(iteratorSeq: Seq[Iterable[NodeIdEdgesMaxId]],
             parallelismLimit: Int,
@@ -291,10 +293,10 @@ object ArrayBasedDirectedGraph {
           val futures = (nodesOutEdges.iterator ++ Iterator(nodesWithNoOutEdges)).map {
             (nodes: Seq[Node]) => futurePool {
               nodes foreach { node =>
+                val fillingInEdgesNode = node.asInstanceOf[FillingInEdgesBiDirectionalNode]
                 val edgeSize = inEdgesSizes(node.id).intValue()
+                fillingInEdgesNode.inEdges = new Array[Int](edgeSize)
                 if (edgeSize > 0) {
-                  val fillingEdgesNode = new FillingInEdgesBiDirectionalNode(node, edgeSize)
-                  table(node.id) = fillingEdgesNode
                   // reset inEdgesSizes, and use it as index pointer of
                   // the current insertion place when adding in edges
                   inEdgesSizes(node.id).set(0)
@@ -330,11 +332,7 @@ object ArrayBasedDirectedGraph {
             nodes => futurePool {
               nodes.foreach {
                 node =>
-                  table(node.id) match {
-                    case fillingInEdgesNode: FillingInEdgesBiDirectionalNode =>
-                      table(node.id) = fillingInEdgesNode.finishingFilling(neighborsSortingStrategy != LeaveUnsorted)
-                    case _ => // no need to finish filling
-                  }
+                  node.asInstanceOf[FillingInEdgesBiDirectionalNode].sortInNeighbors()
               }
             }
           }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
@@ -69,7 +69,7 @@ object NeighborsSortingStrategy extends Enumeration {
 }
 
 object ArrayBasedDirectedGraph {
-  import com.twitter.cassovary.graph.NeighborsSortingStrategy._
+  import NeighborsSortingStrategy._
 
   def apply(iteratorSeq: Seq[Iterable[NodeIdEdgesMaxId]],
             parallelismLimit: Int,
@@ -295,8 +295,8 @@ object ArrayBasedDirectedGraph {
               nodes foreach { node =>
                 val fillingInEdgesNode = node.asInstanceOf[FillingInEdgesBiDirectionalNode]
                 val edgeSize = inEdgesSizes(node.id).intValue()
-                fillingInEdgesNode.inEdges = new Array[Int](edgeSize)
                 if (edgeSize > 0) {
+                  fillingInEdgesNode.createInEdges(edgeSize)
                   // reset inEdgesSizes, and use it as index pointer of
                   // the current insertion place when adding in edges
                   inEdgesSizes(node.id).set(0)
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
@@ -16,9 +16,10 @@ package com.twitter.cassovary.graph
 import com.google.common.annotations.VisibleForTesting
 import com.twitter.cassovary.graph.StoredGraphDir._
 import com.twitter.cassovary.graph.node._
-import com.twitter.cassovary.util.BoundedFuturePool
+import com.twitter.cassovary.util.{SortedArrayOps, BoundedFuturePool}
 import com.twitter.finagle.stats.DefaultStatsReceiver
 import com.twitter.logging.Logger
+import com.twitter.util.Future.when
 import com.twitter.util.{Await, FuturePool, Future}
 import java.util.concurrent.atomic.AtomicInteger
 import scala.collection.mutable
@@ -49,17 +50,43 @@ object NodeIdEdgesMaxId {
 private case class NodesMaxIds(nodesInOnePart: Seq[Node],
                                maxIdInPart: Int, nodeWithOutEdgesMaxIdInPart: Int)
 
+/**
+ * ArrayBasedDirectedGraph can be stored with neighbors sorted or not. Therefore there
+ * are 3 strategies of loading a graph from input:
+ *   `AlreadySorted` - creates a graph with sorted neighbors from sorted input
+ *   `SortWhileReading` - creates a graph with sorted neighbors sorting them
+ *                        while reading
+ *   `LeaveUnsorted` - creates graph with unsorted neighbors (default)
+ */
+object NeighborsSortingStrategy extends Enumeration {
+  type NeighborsSortingStrategy = Value
+
+  val AlreadySorted = Value
+  val SortWhileReading = Value
+  val LeaveUnsorted = Value
+}
+
 object ArrayBasedDirectedGraph {
-  def apply(iterableSeq: Seq[Iterable[NodeIdEdgesMaxId]], parallelismLimit: Int,
-            storedGraphDir: StoredGraphDir):
+  import NeighborsSortingStrategy._
+
+  def apply(iteratorSeq: Seq[Iterable[NodeIdEdgesMaxId]],
+            parallelismLimit: Int,
+            storedGraphDir: StoredGraphDir,
+            neighborsSortingStrategy: NeighborsSortingStrategy = LeaveUnsorted):
   ArrayBasedDirectedGraph = {
-    val constructor = new ArrayBasedDirectedGraphConstructor(iterableSeq, parallelismLimit, storedGraphDir)
+    val constructor = new ArrayBasedDirectedGraphConstructor(iteratorSeq,
+                                                             parallelismLimit,
+                                                             storedGraphDir,
+                                                             neighborsSortingStrategy)
     constructor()
   }
 
   @VisibleForTesting
   def apply(iterable: Iterable[NodeIdEdgesMaxId],
-            storedGraphDir: StoredGraphDir): ArrayBasedDirectedGraph = apply(Seq(iterable), 1, storedGraphDir)
+            storedGraphDir: StoredGraphDir,
+            neighborsSortingStrategy: NeighborsSortingStrategy): ArrayBasedDirectedGraph = {
+    apply(Seq(iterable), 1, storedGraphDir, neighborsSortingStrategy)
+  }
 
   /**
    * Constructs array based directed graph
@@ -67,10 +94,13 @@ object ArrayBasedDirectedGraph {
   private class ArrayBasedDirectedGraphConstructor(
     iterableSeq: Seq[Iterable[NodeIdEdgesMaxId]],
     parallelismLimit: Int,
-    storedGraphDir: StoredGraphDir
+    storedGraphDir: StoredGraphDir,
+    neighborsSortingStrategy: NeighborsSortingStrategy
   ) {
     private lazy val log = Logger.get()
     private val statsReceiver = DefaultStatsReceiver
+
+    private val emptyArray = Array[Int]()
 
     private val futurePool = new BoundedFuturePool(FuturePool.unboundedPool, parallelismLimit)
 
@@ -94,7 +124,7 @@ object ArrayBasedDirectedGraph {
         (table, nodeIdSet) <- markStoredNodes(nodesOutEdges, maxNodeId, storedGraphDir)
         NodesWithNoOutEdgesAndGraphStats(nodesWithNoOutEdges, nodeWithOutEdgesCount, numEdges, numNodes) =
         createNodesWithNoOutEdges(table, nodeIdSet, maxNodeId, storedGraphDir)
-        _ <- Future.when(storedGraphDir == StoredGraphDir.BothInOut) {
+        _ <- when(storedGraphDir == StoredGraphDir.BothInOut) {
           fillMissingInEdges(table, nodesOutEdges, nodesWithNoOutEdges, nodeIdSet, maxNodeId)
         }
       } yield
@@ -148,9 +178,10 @@ object ArrayBasedDirectedGraph {
           id = item.id
           newMaxId = newMaxId max item.maxId
           varNodeWithOutEdgesMaxId = varNodeWithOutEdgesMaxId max item.id
-          val edges = item.edges
-          edgesLength = edges.length
-          val newNode = ArrayBasedDirectedNode(id, edges, storedGraphDir)
+          edgesLength = item.edges.length
+          val edges = if (neighborsSortingStrategy == SortWhileReading) item.edges.sorted else item.edges
+          val newNode = ArrayBasedDirectedNode(id, edges, storedGraphDir,
+            neighborsSortingStrategy != LeaveUnsorted)
           nodes += newNode
         }
         NodesMaxIds(nodes, newMaxId, varNodeWithOutEdgesMaxId)
@@ -224,7 +255,8 @@ object ArrayBasedDirectedGraph {
           if (nodeIdSet(id) == 1) {
             numNodes += 1
             if (table(id) == null) {
-              val node = ArrayBasedDirectedNode(id, ArrayBasedDirectedNode.noNodes, storedGraphDir)
+              val node = ArrayBasedDirectedNode(id, emptyArray, storedGraphDir,
+                neighborsSortingStrategy != LeaveUnsorted)
               table(id) = node
               if (storedGraphDir == StoredGraphDir.BothInOut)
                 nodesWithNoOutEdges += node
@@ -259,14 +291,14 @@ object ArrayBasedDirectedGraph {
           val futures = (nodesOutEdges.iterator ++ Iterator(nodesWithNoOutEdges)).map {
             (nodes: Seq[Node]) => futurePool {
               nodes foreach { node =>
-                val biDirNode = node.asInstanceOf[BiDirectionalNode]
-                val nodeId = biDirNode.id
-                val edgeSize = inEdgesSizes(nodeId).intValue()
-                if (edgeSize > 0)
-                  biDirNode.inEdges = new Array[Int](edgeSize)
-                // reset inEdgesSizes, and use it as index pointer of
-                // the current insertion place when adding in edges
-                inEdgesSizes(nodeId).set(0)
+                val edgeSize = inEdgesSizes(node.id).intValue()
+                if (edgeSize > 0) {
+                  val fillingEdgesNode = new FillingInEdgesBiDirectionalNode(node, edgeSize)
+                  table(node.id) = fillingEdgesNode
+                  // reset inEdgesSizes, and use it as index pointer of
+                  // the current insertion place when adding in edges
+                  inEdgesSizes(node.id).set(0)
+                }
               }
             }
           }.toSeq
@@ -282,8 +314,27 @@ object ArrayBasedDirectedGraph {
               nodes foreach { node =>
                 node.outboundNodes foreach { outEdge =>
                   val index = inEdgesSizes(outEdge).getAndIncrement
-                  table(outEdge).asInstanceOf[BiDirectionalNode].inEdges(index) = node.id
+                  table(outEdge).asInstanceOf[FillingInEdgesBiDirectionalNode].inEdges(index) = node.id
                 }
+              }
+            }
+          }
+          Future.join(futures)
+        }
+      }
+
+      def finishInEdgesFilling(): Future[Unit] = {
+        log.debug("finishing filling")
+        statsReceiver.time("finishing_filling_in_edges") {
+          val futures = nodesOutEdges.map {
+            nodes => futurePool {
+              nodes.foreach {
+                node =>
+                  table(node.id) match {
+                    case fillingInEdgesNode: FillingInEdgesBiDirectionalNode =>
+                      table(node.id) = fillingInEdgesNode.finishingFilling(neighborsSortingStrategy != LeaveUnsorted)
+                    case _ => // no need to finish filling
+                  }
               }
             }
           }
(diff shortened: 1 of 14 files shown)


szymonm:
Hi @pankajgupta 

pls have a look at this.

If the design is fine we can extend it to SharedArrayBased... although reading will be a bit more complicated.

I think we still need to figure out what we really want to sorted neighbors for.

AnishShah:
I can try something for SharedArrayBasedDirectedGraph..

pankajgupta:
@szymonm -- I need to think about this. This design adds the space overhead of an extra class on each adjacency list. What do you think about keeping the knowledge of whether the adjacency lists are sorted or not at the graph level only, and to ensure that the edges are sorted at graph construction time while they are still kept in an array. So the node itself doesn't know whether the adjacency lists are sorted or not but the graph does.

szymonm:
Not a big deal to change it losing some of the 'type' safety. The current implementation will not use sorted versions of operations if array is not wrapped with SortedArrayWrapper, but we can avoid this check. 

I think we should keep the knowledge on the node level but in the class information - like I did using Node with SortedNeighborsNodeOps.

If we don't change it on the node level I see this possibilities:
1) need to add extra parameter for contains/intersect indicating we want to use sorted array based version
2) add class information on Graph level i.e. Graph with SortedEdges, but then we should use contains/intersect implemented in Graph?

The real problem here, I think, is the construction in ArrayBasedDirectedGraph. This code is very messy and I made it even more to feature construction with sorted neighbors...

pankajgupta:
I agree that type safety might be lost, but I guess it can be managed  somewhat? I do think that storage efficiency is key here, so no extra per-node storage.

In the simplest implementation, there is a boolean per graph to keep track of whether the adjs are sorted. Obviously no static type safety here.

To bring about some type safety, can mix in an appropriate trait at graph construction time as you mention.

I am not sure if the messiness of code in ArrayBasedDirectedGraph can be fixed by better code style, or it is just how it is. I am very happy to take your or anyone else's suggestions (perhaps we should do a separate issue for that).

pankajgupta:
@AnishShah -- re SharedArrayBasedDirectedGraph, may I suggest taking a crack at issues 111 and 119?

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedSeq.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+/**
+ * Sequence of sorted elements.
+ * 
+ * Assumes each element may appear in the Seq at most once.
+ */
+trait StrictlySortedSeqLike[T] extends mutable.IndexedSeq[T] {
```
pankajgupta: This trait sounds like SortedSet. Use that instead adding the 'intersect' method?

szymonm: SortedSet uses Ordering/Ordered, so I was afraid of boxing/unboxing, but didn't check it.

szymonm: I see what you mean. Will change it.

Btw, checked speed:
```
import com.twitter.cassovary.util.SortedArrayWrapper

import scala.collection.immutable.SortedSet

def time(a: => Unit): Unit = {
  val t1 = System.currentTimeMillis()
  a
  val t2 = System.currentTimeMillis()
  println("time: " + (t2 - t1))
}

val arr1: Array[Int] = (1 to 1000000).toArray

val arr2: Array[Int] = Range(1, 2000000, 2).toArray

val saw1 = SortedArrayWrapper(arr1)

val saw2 = SortedArrayWrapper(arr2)

time(saw1.intersectSorted(saw2))
val ss1: Set[Int] = SortedSet[Int]() ++ arr1
val ss2 = SortedSet[Int]() ++ arr2
time(ss1.intersect(ss2))
```
SortedArrayWrapper is consistently few orders of magnitude faster.

pankajgupta:
scala> SortedSet[Int]()
res11: scala.collection.immutable.SortedSet[Int] = TreeSet()

The default implementation of the SortedSet trait is TreeSet which is expected to be slower than your array based set for sure. You can make your own array backed concrete implementation of the SortedSet trait. I expect it would be faster than TreeSet() but we will still have to be careful to check that un-necessary boxing/unboxing is not occurring on every element.

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedSeq.scala
szymonm: If I wanted SortedArrayWrapper extends Seq with SortedSetLike, I need it to implement: contains, + (adding element) and -. Then I can override intersect, but:
```
def intersect(that: GenSet[A]): This
```
thus, if I override it, this method will be called for each GenSet not only SortedSets. This makes an implementation in SortedArrayWrapper not concise with the definition.

pankajgupta: yeah, that's a lot of methods to override that are unnecessary for our purpose.

btw, do you still need this wrapper when moving to a graph-based design rather than a node-based design?

szymonm: I need some of the methods here. Probably not the wrapper itself.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
```diff
@@ -51,9 +51,12 @@ private case class NodesMaxIds(nodesInOnePart: Seq[Node],
 
 object ArrayBasedDirectedGraph {
   def apply(iteratorSeq: Seq[() => Iterator[NodeIdEdgesMaxId]], parallelismLimit: Int,
-            storedGraphDir: StoredGraphDir):
+            storedGraphDir: StoredGraphDir, neighborsSorted: Boolean = false):
   ArrayBasedDirectedGraph = {
-    val constructor = new ArrayBasedDirectedGraphConstructor(iteratorSeq, parallelismLimit, storedGraphDir)
+    val constructor = new ArrayBasedDirectedGraphConstructor(iteratorSeq,
```
pankajgupta: do you remember why we have this extra private class whose apply() is called to produce ArrayBasedDirectedGraph (as opposed to just having the code in the constructor inline here)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -13,6 +13,8 @@
  */
 package com.twitter.cassovary.graph
 
+import com.twitter.cassovary.util.SortedArrayOps
+
```
pankajgupta: delete extra line

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -215,6 +217,20 @@ trait Node {
   }
 
   /**
+   * @return Intersection of `dir` neighbors with `nodeIds`.
+   */
+  def intersect(dir: GraphDir, nodeIds: Seq[Int]): Seq[Int] = {
+    dir match {
```
pankajgupta: use existing method neighborIds()

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -273,9 +289,15 @@ trait Node {
...
-  def apply(nodeId: Int, out: Seq[Int], in: Seq[Int]) = new Node {
-    val id = nodeId
-    def inboundNodes() = in
-    def outboundNodes() = out
+
+  def apply(nodeId: Int, out: Seq[Int], in: Seq[Int]): Node = {
+    new SeqBasedNode(nodeId, in, out)
+  }
+
+  def withSortedNeighbors(nodeId: Int, in: Array[Int], out: Array[Int]) = {
```
pankajgupta: the order of 'in' and 'out' in parameter list is different here which could be a potential source of confusion/bugs

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -273,9 +289,15 @@ trait Node {
...
+    new SeqBasedNode(nodeId, in, out) with SortedNeighborsNodeOps
   }
 }
+
+class SeqBasedNode private[graph] (val id: Int, val inboundNodes: Seq[Int], val outboundNodes: Seq[Int])
```
pankajgupta: not sure I like this. SeqBasedNode is really just the same as Node (with no added or reduced functionality), so may be we should just make the trait Node to be an abstract class?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SortedNeighborsNodeOps.scala
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2014 Twitter, Inc.
```
pankajgupta: please use 2015 for all new files

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
+package com.twitter.cassovary.util
+
+import java.util
+
```
pankajgupta: delete new line and also do import java.{util => jutil} to be a bit clearer.

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+object SortedArrayOps {
+  /**
+   * Finds element in the SortedSeq using provided ordering.
```
pankajgupta: update comment. Not clear what 'ordering' is.

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+   *
+   * Assumes that the collection is sorted using this ordering.
+   */
+  def findUsingOrdering(array: Array[Int], t: Int): Boolean = {
```
pankajgupta: rename to 'exists' ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+      case idx if idx < 0 => false
+      case idx => true
+    }
+  }
+
+  /**
+   * Linear intersection of two sorted arrays.
+   */
+  def intersectSorted(array: Array[Int], that: Array[Int]): Array[Int] = {
+    val result = Array.ofDim[Int](array.length + that.length)
```
pankajgupta: should be min not +

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+      } else {
+        thatIdx += 1
+      }
+    }
+
+    util.Arrays.copyOfRange(result, 0, resultIdx)
+  }
+
+  /**
+   * Linear intersection of two sorted arrays.
```
pankajgupta: How is this method different? Is this a merge?

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+    def putInDest(x: Int): Unit = {
+      result(resultIdx) = x
+      resultIdx += 1
+    }
+
+    while (arrayIdx < array.length || thatIdx < that.length) {
+      if (arrayIdx == array.length) {
+        putInDest(that(thatIdx))
+        thatIdx += 1
+      } else if (thatIdx == array.length) {
```
pankajgupta: did you mean that.length here instead of array.length?

Also, please check that tests cover the conditions.

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+          putInDest(array(arrayIdx))
+          arrayIdx += 1
+        } else {
+          putInDest(that(thatIdx))
+          thatIdx += 1
+        }
+      }
+    }
+
+    util.Arrays.copyOfRange(result, 0, resultIdx)
```
pankajgupta: seems like scala/collection/mutable/ArrayOps.scala uses Array.copy which is a system call to System.arraycopy . Might want to use that?

cassovary-core/src/test/scala/com/twitter/cassovary/util/SortedArrayOpsTest.scala
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.util
+
+import org.scalatest.{Matchers, WordSpec}
+
+class SortedArrayOpsTest extends WordSpec with Matchers {
```
pankajgupta: s/Test/Spec

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
+package com.twitter.cassovary.util
+
+import java.util
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
```
pankajgupta: is this required?

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
```diff
@@ -0,0 +1,85 @@
...
+
+object SortedArrayOps {
+  /**
+   * Finds element in the SortedSeq using provided ordering.
+   *
+   * Assumes that the collection is sorted using this ordering.
+   */
+  def findUsingOrdering(array: Array[Int], t: Int): Boolean = {
+    util.Arrays.binarySearch(array, t) match {
```
pankajgupta: Check scala.collection.Searching -- wondering if we should use IndexedSeq[Int] instead of Array[Int] here.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SortedNeighborsNodeOps.scala
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.graph
+
+import com.twitter.cassovary.util.SortedArrayOps
+
+trait SortedNeighborsNodeOps {
```
pankajgupta: this is nice. Some comments here would be useful to the reader.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SortedNeighborsNodeOps.scala
```diff
@@ -0,0 +1,28 @@
...
+  self: Node =>
+
+  override protected def containsNode(nodeIds: Seq[Int], queryNodeId: Int): Boolean = {
```
pankajgupta: we probably want nodeIds to be of type IndexedSeq or Array (instead of converting nodeIds to an array).

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
```diff
@@ -225,7 +228,7 @@ object ArrayBasedDirectedGraph {
           if (nodeIdSet(id) == 1) {
             numNodes += 1
             if (table(id) == null) {
-              val node = ArrayBasedDirectedNode(id, ArrayBasedDirectedNode.noNodes, storedGraphDir)
+              val node = ArrayBasedDirectedNode(id, Array[Int](), storedGraphDir, neighborsSorted)
```
pankajgupta: I think Array[Int]() creates a new empty object everytime. That was the idea to have noNodes.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
```diff
@@ -149,9 +153,8 @@ object ArrayBasedDirectedGraph {
           id = item.id
           newMaxId = newMaxId max item.maxId
           varNodeWithOutEdgesMaxId = varNodeWithOutEdgesMaxId max item.id
-          val edges = item.edges
-          edgesLength = edges.length
-          val newNode = ArrayBasedDirectedNode(id, edges, storedGraphDir)
+          edgesLength = item.edges.length
```
pankajgupta: we want to sort the edges if we are asked to (this was the original motivation). i.e., neighborsSorted is actually sortNeighbors. If neighbors is already sorted, well it is okay to sort it again if we are asked to. (If neighbors is already sorted and we do not want to sort them again, we will deal with the case when needed).

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
```diff
@@ -292,10 +296,26 @@ object ArrayBasedDirectedGraph {
...
       }
 
+      def finishInEdgesFilling(): Future[Unit] = {
+        log.debug("finishing filling")
+        statsReceiver.time("finishing_filling_in_edges") {
+          val futures = nodesOutEdges.map {
+            nodes => futurePool {
+              nodes.foreach {
+                node =>
+                  table(node.id) = node.asInstanceOf[FillingInEdgesBiDirectionalNode].finishingFilling()
```
pankajgupta: again sort the edges if sortNeighbors is true.

pankajgupta:
Design LGTM otherwise. Might want to add a test at the graph level.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
szymonm: Having the code in constructor makes it hard to override.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
szymonm: No, because of other methods. It is a bit more general. Consider:
```
class SetBasedNode(val id, val in: Set[Int], val out: Set[Int]) extends Node {
  def inboundNodes() = in.toSeq // slower
  def isInboundNode(node: Int) = ... // faster
  ...
}
```
You cannot do this using abstract class Node based on seqs.

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
szymonm: The methods there use Ordering, so I'm afraid they use boxing as well.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SortedNeighborsNodeOps.scala
szymonm: nodeIds are by default still a Seq. And IndexedSeq still needs toArray `conversion`.

containsNodes works also when neighbors are not sorted. so Seq is used.

If a Seq is an array than toArray does nothing.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
szymonm: neighborsSorted means that neighbors are already sorted in the input. Is this ok?

We have to sort inNodes at the end.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
pankajgupta: My question was different -- as currently defined, the purpose of SeqBasedNode is not clear at all (to me). Seems like it doesn't provide anything new or different than Node itself. i.e., whatever you do with SeqBasedNode you can directly do based on Node.

cassovary-core/src/main/scala/com/twitter/cassovary/util/SortedArrayOps.scala
pankajgupta: that's right. Please add a comment then that we are writing our own to not have boxing/unboxing overhead of scala.collection.Searching

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
szymonm: It's only a constructor, not any logic at all.

I have it only in order not to write:
```
new Node {
  val id = nodeId
  override def inboundNodes() = in
  overrided def outboundNodes() = out
}
```

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
szymonm: Shall we allow also to read graph that is not sorted and sort it while reading?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SortedNeighborsNodeOps.scala
pankajgupta: If you don't do an 'override', then the compiler will choose this implementation based on the method signature which could be an array here. i.e.,
protected def containsNode(nodeIds: Array[Int], queryNodeId: Int): Boolean = ....

then the compiler will choose this and you don't have to write the code for or do an .toArray at runtime (I think)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
pankajgupta: No, my original intention was for the code to sort it if it is given in unsorted order so we want a different variable 'sortNeighbors' which the code checks and when true sorts every adjacency list.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
pankajgupta: perhaps add a single line comment to make it clear

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
pankajgupta: ah, the answer is yes to your question.

pankajgupta:
is this waiting on me? if so, what about sorting the edges if asked by the parameter?

szymonm:
Nope, I have to still add sorting edges. Pls wait till tomorrow evening (PST).

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SortedNeighborsNodeOps.scala
szymonm: Checked this and it does not work as you said. If I use
```
protected def containsNode(nodeIds: Array[Int], queryNodeId: Int): Boolean
```
the method is not called when I call constainsNode on inboundNodes() that are an array.

Also if I omit override, compiler raises error: 
```
SeqBasedNode with com.twitter.cassovary.graph.SortedNeighborsNodeOps> inherits conflicting members
```

pankajgupta: Interesting. it works in my repl:

scala> trait A { protected def f(s: Seq[Int]): Boolean = true }
defined trait A

scala> class A2 extends A { protected def f(a: Array[Int]) = false 
     | println(f(Array(1)))
     | }
defined class A2

scala> val a2 = new A2
false
a2: A2 = A2@4f2c9ba6

scala> object A3 extends A2 { println("In Seq:" + f(Seq(1,2))); println("In Array:" + f(Array(0))) }
defined module A3

scala> A3
false
In Seq:true
In Array:false
res5: A3.type = A3$@4a9f80d3

szymonm: I didn't know about that.

But we still have a problem here. I used as before but:
```
object A3 extends A2 {
  val arr: Seq[Int] = Array(0)
  println("In Seq:" + f(Seq(1,2))); println("In Array:" + f(arr))
}
```
and obtained:
```
// Exiting paste mode, now interpreting.

defined object A3

scala> A3
false
In Seq:true
In Array:true
res0: A3.type = A3$@51b69452
```

szymonm: Do you think we can fix it?

szymonm: I fixed all but this in the last commit.

pankajgupta: It is clear from these examples that the choice of method is based on
compile-time type... No good for us.

szymonm:
Any toughts why Travis is not running the last commit?

pankajgupta:
Not sure if anyone knows, but it does do that sometimes... @caniszczyk any ideas?

szymonm:
Do you think it's ok but that? I'll squash the commits then.

AnishShah:
@szymonm @pankajgupta I guess, because of merge conflict ?

 http://docs.travis-ci.com/user/pull-requests/#My-Pull-Request-isn't-being-built

pankajgupta:
just to be sure, would be great to see before merging if these changes have degraded the speed of construction of array based directed graph in the leave unsorted case

szymonm:
Tested on wiki-Vote and facebook graphs in cassovary-bechmarks. Running times are very close (all four about 1s 200millis). Do you need larger tests?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/ArrayBasedDirectedNode.scala
```diff
@@ -23,15 +23,16 @@ object ArrayBasedDirectedNode {
...
    */
-  def apply(nodeId: Int, neighbors: Array[Int], dir: StoredGraphDir) = {
+  def apply(nodeId: Int, neighbors: Seq[Int], dir: StoredGraphDir, sortedNeighbors: Boolean = false) = {
     dir match {
       case StoredGraphDir.OnlyIn | StoredGraphDir.OnlyOut | StoredGraphDir.Mutual =>
-        UniDirectionalNode(nodeId, neighbors, dir)
+        UniDirectionalNode(nodeId, neighbors, dir, sortedNeighbors)
       case StoredGraphDir.BothInOut =>
-        BiDirectionalNode(nodeId, neighbors)
+        BiDirectionalNode(nodeId, neighbors, sortedNeighbors)
```
pankajgupta: I was suggesting changing this to creating a FillingInEdgesBiDirectionalNode and not a BiDirectionalNode. This removes one unnecessary class instance creation "table(node.id) = fillingEdgesNode" in ArrayBasedDirectedGraph above. This also will require modification to finishInEdgesFilling() to not do a pattern match.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
```diff
@@ -291,10 +323,30 @@ object ArrayBasedDirectedGraph {
...
+          }
+          Future.join(futures)
+        }
+      }
+
       for {
         inEdgesSizes <- findInEdgesSizes(nodesOutEdges, nodeIdSet, numNodes)
         _ <- instantiateInEdges(inEdgesSizes)
         _ <- populateInEdges(inEdgesSizes)
+        _ <- when(neighborsSortingStrategy != LeaveUnsorted) (finishInEdgesFilling())
```
pankajgupta: one effect of this condition is of course that under this sorting strategy, nodes are of type FillingInEdgesBiDirectionalNode are left as it is (which is okay with me as table[] is of type Array[Node] but you had pointed out earlier that this was not ideal)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/ArrayBasedDirectedNode.scala
szymonm: But then we need casting to FillingInEdges... (from Node) when initializing array, filing edges and another casting when finishing filling. I thought we wanted to avoid this.

pankajgupta: Casting is fine. We have casting even now while filling edges. 
I guess if we really wanted to avoid casting, we should create a separate tempTable: Array[FillingInEdgesBiDirectionalNode] and assign to table once edges have been filled.

szymonm: We would still need casting because cannot do:
```
val a: Array[BiDirectionalNode] = new Array[FillingInEdgesBiDirectionalNode]
```
Ok, will change it.

pankajgupta: may be just let a be of type FillingInEdgesBiDirectionalNode ?

szymonm:
Is this what you meant?

pankajgupta:
yes, this looks good. Minor comments coming up next.

PR #135, opened 2015-01-27
Array Based Dynamic Graph

plofgren:
I've implemented a dynamic directed graph, using an ArrayBuffer of IntArrayList (from the fastutil library). If n nodes are used, O(n) objects are created, independent of the number of edges.

Comparison to the current class: Note that the current dynamic graph class SynchronizedDynamicGraph uses a ConcurrentHashMap to store nodes, which has more overhead than an ArrayBuffer when the nodes are mostly sequential.  Also, each node stores an ArrayBuffer[Int], which I belive boxes the ints into objects, creating overhead relative to fastutil's IntArrayList.  It also has non-trivial synchronization overhead; in particular when iterating over the neighbors of a node, the neighbors are first copied into a new array, then an iterator to the new array is returned.  The current class is better when the node Ids are very non-sequential or when automatic synchronization is needed, but otherwise I believe the new class is more efficient.  In a very informal comparison on a graph with several hundred million edges, the currrent class wouldn't load using 30GB of heap, while the new class fit the graph in 7.3 GB of RAM.

@pankajgupta Could you take a look at this, or point me to someone else? 

Thanks!

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
@@ -37,7 +37,7 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
       getOrCreateNode(id)
       nodeData.edges map getOrCreateNode
       storedGraphDir match {
-        case OnlyOut => outboundLists(id).addAll(IntArrayList.wrap(nodeData.edges))
+        case OnlyOut | Mutual => outboundLists(id).addAll(IntArrayList.wrap(nodeData.edges))
         case OnlyIn => inboundLists(id).addAll(IntArrayList.wrap(nodeData.edges))
         case BothInOut => nodeData.edges map { addEdgeAllowingDuplicates(id, _) } // Duplicates shouldn't exist, but allow them for efficiency.
       }
@@ -69,18 +69,15 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
     * For efficiency, we don't store Nodes, but create them on the fly as needed.
     */
   class IntListNode(override val id: Int,
-                    val outboundList: Option[IntArrayList],
-                    val inboundList: Option[IntArrayList])
+                    val outboundList: Option[Seq[Int]],
+                    val inboundList: Option[Seq[Int]])
       extends DynamicNode {
-    override def outboundNodes(): Seq[Int] = outboundList match {
-      case Some(list) => FastUtilUtils.intArrayListToSeq(list)
-      case None => Seq.empty[Int]
-    }
+    override def outboundNodes(): Seq[Int] = outboundList.getOrElse(Seq.empty)
 
-    override def inboundNodes(): Seq[Int] = inboundList match {
-      case Some(list) => FastUtilUtils.intArrayListToSeq(list)
-      case None => Seq.empty[Int]
-    }
+    override def inboundNodes(): Seq[Int] = storedGraphDir match {
+      case Mutual => outboundNodes()
+      case _ => inboundList.getOrElse(Seq.empty)
+      }
 
     def addOutBoundNodes(nodeIds: Seq[Int]): Unit = {
       //For future optimization, we could check if we are only storing outbound
@@ -110,7 +107,9 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
    */
   override def getNodeById(id: Int): Option[DynamicNode] =
     if (nodeExists(id)) {
-      Some(new IntListNode(id, outboundListOption(id), inboundListOption(id)))
+      Some(new IntListNode(id,
+        outboundListOption(id) map FastUtilUtils.intArrayListToSeq,
+        inboundListOption(id) map FastUtilUtils.intArrayListToSeq))
     } else {
       None
     }
@@ -140,6 +139,9 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
   override def removeEdge(srcId: Int, destId: Int): (Option[Node], Option[Node]) = {
     outboundListOption(srcId) map {list => list.rem(destId)}
     inboundListOption(destId) map {list => list.rem(srcId)}
+    if (storedGraphDir == Mutual) {
+      outboundListOption(destId) map {list => list.rem(srcId)}
+    }
     (getNodeById(srcId), getNodeById(destId))
   }
 
@@ -156,6 +158,10 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
         list.add(k)
     outboundListOption(srcId) map { addIfMissing(_, destId) }
     inboundListOption(destId) map { addIfMissing(_, srcId) }
+    if (storedGraphDir == StoredGraphDir.Mutual) {
+      // Add mutual edge also
+      outboundListOption(destId) map { addIfMissing(_, srcId) }
+    }
   }
 
   /**
@@ -166,11 +172,13 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
   def addEdgeAllowingDuplicates(srcId: Int, destId: Int): Unit = {
     getOrCreateNode(srcId)
     getOrCreateNode(destId)
-    def addIfMissing(list: IntArrayList, id: Int): Unit =
-      if (!list.contains(destId))
-        list.add(destId)
+
     outboundListOption(srcId) map { _.add(destId) }
     inboundListOption(destId) map { _.add(srcId) }
+    if (storedGraphDir == StoredGraphDir.Mutual) {
+      // Add mutual edge also
+      outboundListOption(destId) map { _.add(srcId) }
+    }
   }
 
   /**
@@ -189,7 +197,8 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
       if (StoredGraphDir.isOutDirStored(storedGraphDir)) {
         addIdToList(outboundLists)
       }
-      if (StoredGraphDir.isInDirStored(storedGraphDir)) {
+      if (StoredGraphDir.isInDirStored(storedGraphDir) &&
+        storedGraphDir != StoredGraphDir.Mutual) {
         addIdToList(inboundLists)
       }
     }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
@@ -46,7 +46,7 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
 
   def this(iteratorSeq: Seq[() => Iterator[NodeIdEdgesMaxId]],
            storedGraphDir: StoredGraphDir) {
-    this((iteratorSeq map { _()} ).iterator.flatten, storedGraphDir)
+    this((iteratorSeq.view flatMap { _()}).iterator, storedGraphDir)
   }
 
   /* Returns an option which is non-empty if outbound list for id  is non-null. */
--- a/cassovary-core/src/test/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraphSpec.scala
+++ b/cassovary-core/src/test/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraphSpec.scala
@@ -5,6 +5,9 @@ import com.twitter.cassovary.graph.StoredGraphDir._
 import org.scalatest.matchers.ShouldMatchers
 
 class ArrayBasedDynamicDirectedGraphSpec extends WordSpec with ShouldMatchers with GraphBehaviours {
+  val graphDirections = List(StoredGraphDir.OnlyIn, StoredGraphDir.OnlyOut, StoredGraphDir.BothInOut,
+                             StoredGraphDir.Mutual) // Bipartitie is not supported
+
   def builder(iteratorFunc: () => Iterator[NodeIdEdgesMaxId],
               storedGraphDir: StoredGraphDir) =
     new ArrayBasedDynamicDirectedGraph(iteratorFunc(), storedGraphDir)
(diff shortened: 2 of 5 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
```diff
@@ -61,4 +61,6 @@ object FastUtilUtils {
 
   def newInt2IntOpenHashMap(): mutable.Map[Int, Int] =
     new Int2IntOpenHashMap().asInstanceOf[jutil.Map[Int, Int]]
+
+  def intArrayListToSeq(list: IntArrayList): Seq[Int] = list map { _.toInt }
```
pankajgupta: list.toIntArray() will also work, but this is fine too.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
+package com.twitter.cassovary.graph
+
+import com.twitter.cassovary.graph.node.DynamicNode
```
pankajgupta: please arrange imports in alphabetical order

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+ * (from the fastutil library).  If n nodes are used, O(n) objects are created,
+ * independent of the number of edges.
+ *
+ * For efficiency, it's recommended that ids be sequentially numbered, as
+ * an internal array is stored which is longer than the maximum id seen.
+ *
+ * This class is not threadsafe.  If multiple thread access it simultaneously
+ * and at least one mutates it, synchronization should be used.
+ */
+class ArrayBasedDynamicDirectedGraph(override val storedGraphDir: StoredGraphDir)
```
pankajgupta: don't need 'override' specifier.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+    extends DynamicDirectedGraph {
+  // outboundLists(id) contains the outbound neighbors of the given id,
+  // or null if the id is not in this graph.
+  // If we aren't storing outbound neighbors, outboundLists will always remain size 0.
+  private val outboundLists = new ArrayBuffer[IntArrayList]
+  // (See above note on outboundLists)
+  private val inboundLists = new ArrayBuffer[IntArrayList]
+
+  private var nodeCount_ = 0
```
pankajgupta: scala doesn't use the _ convention too much.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+        case OnlyOut => outboundLists(id).addAll(IntArrayList.wrap(nodeData.edges))
+        case OnlyIn => inboundLists(id).addAll(IntArrayList.wrap(nodeData.edges))
+        case BothInOut => nodeData.edges map { addEdgeAllowingDuplicates(id, _) } // Duplicates shouldn't exist, but allow them for efficiency.
+      }
+    }
+  }
+
+  def this(iteratorSeq: Seq[() => Iterator[NodeIdEdgesMaxId]],
+           storedGraphDir: StoredGraphDir) {
+    this((iteratorSeq map { _()} ).iterator.flatten, storedGraphDir)
```
pankajgupta: can use flatMap: 
iteratorSeq flatMap { _() }

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * A dynamic directed graph, implemented using an ArrayBuffer of IntArrayList
+ * (from the fastutil library).  If n nodes are used, O(n) objects are created,
+ * independent of the number of edges.
+ *
+ * For efficiency, it's recommended that ids be sequentially numbered, as
+ * an internal array is stored which is longer than the maximum id seen.
+ *
+ * This class is not threadsafe.  If multiple thread access it simultaneously
```
pankajgupta: Perhaps also add for clarity that all operations are done in a single thread.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+    else
+      Option(inboundLists(id)) // If list is null, convert to None
+
+  /** Wraps outboundList and inboundList in a node structure.  This is an inner class
+    * because mutations will affect other nodes of the graph if both inbound
+    * and outbound neighbor lists are stored.
+    *
+    * For efficiency, we don't store Nodes, but create them on the fly as needed.
+    */
+  class IntListNode(override val id: Int,
```
pankajgupta: the recommended way to do what you are doing IIUC (to expose only a DynamicNode type to the outside world) is to declare a private object like this:

private object IntListNode {

def apply(id: Int, outboundList: ..., inboundList: ...) = new DynamicNode {
....
}
Besides being cleaner, you are saving on not having to pay for storage of outboundList and inboundList in the created instance of IntListNode

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+                    val outboundList: Option[IntArrayList],
+                    val inboundList: Option[IntArrayList])
+      extends DynamicNode {
+    override def outboundNodes(): Seq[Int] = outboundList match {
+      case Some(list) => FastUtilUtils.intArrayListToSeq(list)
```
pankajgupta: Bear in mind that random access (e.g., for a random walk) would first require traversal of the whole intArrayList to convert the list to a Seq rather than an O(1) direct access. What you want to do I think is to define a new wrapper class like this:
class IntArrayListWrapper(l: IntArrayList) extends IndexedSeq[Int] {
def apply(idx: Int): Int = l.getInt(idx)
def length: Int = l.size
}

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
pankajgupta: this will be O(n) btw

pankajgupta:
Consider adding benchmark in cassovary-benchmark subproject as well to demo the performance of this.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * This class is not threadsafe.  If multiple thread access it simultaneously
+ * and at least one mutates it, synchronization should be used.
+ */
+class ArrayBasedDynamicDirectedGraph(override val storedGraphDir: StoredGraphDir)
+    extends DynamicDirectedGraph {
+  // outboundLists(id) contains the outbound neighbors of the given id,
+  // or null if the id is not in this graph.
+  // If we aren't storing outbound neighbors, outboundLists will always remain size 0.
+  private val outboundLists = new ArrayBuffer[IntArrayList]
```
szymonm: You can also consider using http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/objects/ObjectArrayList.html instead ArrayBuffer

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -0,0 +1,206 @@
...
+
+    def removeOutBoundNode(nodeId: Int): Unit = removeEdge(id, nodeId)
+
+    def removeInBoundNode(nodeId: Int): Unit =  removeEdge(nodeId, id)
+  }
+
+  /**
+   * Determine if a node with the given id exists in this graph.
+   */
+  def nodeExists(id: Int): Boolean =
```
szymonm: Throws NPE when id > outboundLists.length

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
```diff
@@ -61,4 +61,6 @@ object FastUtilUtils {
 
   def newInt2IntOpenHashMap(): mutable.Map[Int, Int] =
```
szymonm: Have a look here. I'm creating fastutils collection that is only an implementation of scala's built-in collection (mutable.Map). If you use something like this, you can have 
```
private val outboundLists: mutable.IndexedSeq[mutable.IndexedSeq[Int]] = new ArrayBuffer()
```
IMO, it's more convenient to hide fastutils api and use scala's that is easier in use (ex. map, find and so on), esp. in tests.

This works using the fact that IntArrayList implements java.util.List that can be converted to scala.collection.mutable.Seq using scala.collection.JavaConversions._.

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
szymonm: You don't need to use this at all (see my previous comment)

cassovary-core/src/test/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraphSpec.scala
```diff
@@ -0,0 +1,76 @@
+package com.twitter.cassovary.graph
+
+import org.scalatest.WordSpec
+import com.twitter.cassovary.graph.StoredGraphDir._
+import org.scalatest.matchers.ShouldMatchers
+
+class ArrayBasedDynamicDirectedGraphSpec extends WordSpec with ShouldMatchers with GraphBehaviours {
```
szymonm: See here for the structure of a test:
http://doc.scalatest.org/2.2.1/#org.scalatest.WordSpec

To obtain nice class spec when running test, it should be sth like 
```
"An ArrayBasedDynamicDirectedGraph" should {
   "add new nodes" in {
   }
  "delete edges" in {
  }
}
```

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
plofgren: Good point about O(n)!  I've now changed it to
def intArrayListToSeq(list: IntArrayList): Seq[Int] = list.elements()
which should be O(1).  I couldn't find an immutable.Seq wrapper in the standard library.  This has the downside that technically a client could cast the result to WrappedArray[Int] and then mutate it, but clients who do that should expect mal-formed graphs

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Done

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: I agree it isn't needed, but I thought it was good style. I'll remove it.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Is there a better name for _nodeCount? (I just realized that http://docs.scala-lang.org/style/naming-conventions.html#accessorsmutators puts the underscore before the name).

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: I considered that, but I think it would convert the iterators into a single Seq (reading all the data from all the iterators into memory), and I thought that might give worse performance than constructing the graph as we iterate over the input.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Sounds good.  I may add threads in the future if the performance seems important in practice.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: This wasn't that clear, but I meant to say that it is an inner class because it needs a reference to this (the graph) in the mutator methods (they call addEdge on the enclosing graph).  If I change it to a static class as you're suggesting, I would need to pass in a reference to this (which I'm happy with).  Why would your alternative avoid storing outboutList and inboundList?  Wouldn't they still be stored in a closure somewhere?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Nice, this is the read-only wrapper I can use.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Do you think it is more efficient than ArrayBuffer?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: I think outboundListOption already handles the case when id > outboundLists.length.  I just added
    graph.nodeExists(1000000) shouldEqual false
to one of the tests (where graph is a small graph), and it passed.

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
plofgren: I agree that's easier to use IndexedSeq, but I'm slightly afraid that if I used map, etc. on IndexedSeq I'll accidentally do an O(n) copy when I'm not expecting it.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
szymonm: No idea.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
szymonm: You're right.

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
szymonm: 1) That's Scala's charm.
2) Should be fine as long as you use val.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
pankajgupta: I think none of the iterators should be called -- easy to test in repl.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
pankajgupta: at least in ArrayBasedDirectedGraph, single thread performance was abysmal. From our past experience, reading graph is a time consuming operation and so it might pay to do it in multiple threads

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
pankajgupta: here you always pay for the storage of the two option classes which won't be the case otherwise.

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
pankajgupta: @szymonm in your suggestion the inner mutable.IndexedSeq will have unboxing/boxing, yes?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Good idea to just test it.  If I execute the following script:
```scala
def makeIterator(name: String): Iterator[Int] = new Iterator[Int] {
  var current = 0
  def next(): Int =  {
    println("iterator " + name + " returning " + current)
    current += 1
    current - 1
  }
  def hasNext(): Boolean = (current < 5)
}
val iteratorSeq: Seq[() => Iterator[Int]] = Seq(
    () => makeIterator("one"),
    () => makeIterator("two"),
    () => makeIterator("three"))

val flatMapResult = iteratorSeq flatMap { _()}
```
Then it prints 15 lines of the form "iterator _ returning _"
Wheras if I replace the last line with
`val moreAwkwardResult = (iteratorSeq map { _()} ).iterator.flatten`
then it prints nothing.

plofgren: Perhaps a clearer alternative to what I have is 
`iteratorSeq.view flatMap { _()}`

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: If we want to avoid the storage of one of the two option classes, I think we would need to not access it inside the relevant closure, which would make the code awkward (I'm not sure how to do that without essentially writing two different Node classes). More precicely, in your code
`new DynamicNode {
....
}`
I assume the "...." contains references to both outboundList and inboundList, inless there are separate calls to `new Dynamic Node {...}` for each case.  Since Nodes aren't being stored anyway, I'm inclinded to leave it as is unless you think it's important.

plofgren:
I probably won't get to the benchmark for a few days, but I'll post once I have it, and I'll make a call then about multi-threading.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
pankajgupta: You are right. We will need to avoid the references inside the {...}. At the minimum, you can do upfront the logic of converting to Seq like:
val o: Seq[Int] = outboundList.map{fastUtilUtils.intArrayListToSeq()) getOrElse(Seq.empty[Int])

and then use 'i' or 'o' inside the new DynamicNode {...} . Will this be better? Might not be a big saving regardless, so up to you.

pankajgupta:
ok regarding benchmark

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Doing the conversion to Seq outside IntListNode is actually a bit cleaner, so I did that.

plofgren:
I needed a mutable undirected graph, so I just added code to support undirected graphs (i.e. stored direction Mutual).
@pankajgupta @szymonm Please comment on the new commit when you get a chance.

plofgren:
I actually didn't realize until today that "Mutual" meant undirected and was starting to write my own UndirectedGraph wrapper class when I noticed it.  Should I add a comment to StoredGraphDir along the lines of
`Mutual, // In a mutual graph, the outbound and inbound neighbors of each node are equal, and space is typically saved by only storing the outbound neighbors of each node`
?

pankajgupta:
LGTM. Yes, Mutual has that intention, but it will be better for there to be a barebones UndirectedGraphWrapper class for readability and to avoid the surprise you ran into. 
So let's close this PR and please feel free to take up that into a new PR.

cassovary-core/src/main/scala/com/twitter/cassovary/util/FastUtilUtils.scala
szymonm: I cannot answer, but you are right that wrapping the fastutils collection is expensive. See #142

PR #137, opened 2015-01-28
Remove the unnecessary "() =>" in `def iteratorSeq` 

AnishShah:
I made an abstract class as @szymonm said.

```
abstract class NodeStream extends Iterable[NodeIdEdgesMaxId] with Closeable {..}
```
Then, I used it in AdjacencyListGraphReader.scala & ListOfEdgesGraphReader.scala.

I implemened the `close()` method to close the file.

Is this correct? I haven't removed "() => " in `def iteratorSeq`.  I will do it after reviews on this.

PR #138, opened 2015-01-28
Degree centrality

bmckown:
This PR is the beginning of building out the analysis framework to allow more metrics to be computed on the graph.  We start with the simplest, in-degree and out-degree centrality, and will move to more complicated centrality algorithm and iterative algorithms such as HITS.

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -18,5 +18,9 @@ import com.twitter.cassovary.graph.DirectedGraph
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
   val centrality = new Array[Double](graph.maxNodeId + 1)
 
+  /**
+   * Normalize the values
+   * @return
+   */
   def normalize: Array[Double] = centrality.map(c => c / (graph.maxNodeId - 1))
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -16,6 +16,11 @@ package com.twitter.cassovary.algorithms.centrality
 import com.twitter.cassovary.graph.{Node, DirectedGraph}
 import collection.immutable.IndexedSeq
 
+/**
+ * Centrality algorithm abstraction that requires a directed graph {@code graph} be passed upon
+ * instantiation.
+ * @param graph A directed graph
+ */
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
 
   val centralityValues = new Array[Double](graph.maxNodeId + 1)
@@ -23,9 +28,13 @@ abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
   /**
    * Get the specified node's centrality value
    * @param n Node
-   * @return
+   * @return Centrality value for the n-th node
    */
   def apply(n: Node): Double = centralityValues(n.id)
 
+  /**
+   * Run the centrality calculation and update {@code centralityValues}
+   * @return The centrality values of the graph
+   */
   def recalculate: IndexedSeq[Double]
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -14,6 +14,7 @@
 package com.twitter.cassovary.algorithms.centrality
 
 import com.twitter.cassovary.graph.{Node, DirectedGraph}
+import collection.immutable.IndexedSeq
 
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
 
@@ -24,7 +25,7 @@ abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
    * @param n Node
    * @return
    */
-  def apply(n: Node): Double = centrality(n.id)
+  def apply(n: Node): Double = centralityValues(n.id)
 
-  def centrality: Array[Double]
+  def recalculate: IndexedSeq[Double]
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -1,0 +1,10 @@
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.DirectedGraph
+
+/**
+ * Created by bmckown on 1/28/15.
+ */
+abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+  val centrality = new Array[Double](graph.maxNodeId + 1)
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -13,14 +13,16 @@
  */
 package com.twitter.cassovary.algorithms.centrality
 
-import com.twitter.cassovary.graph.DirectedGraph
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
 
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
-  val centrality = new Array[Double](graph.maxNodeId + 1)
+  def centrality: Array[Double]
 
   /**
    * Normalize the values
    * @return
    */
   def normalize: Array[Double] = centrality.map(c => c / (graph.nodeCount - 1))
+
+  def apply(n: Node): Double = centrality(n.id)
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -17,4 +17,6 @@ import com.twitter.cassovary.graph.DirectedGraph
 
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
   val centrality = new Array[Double](graph.maxNodeId + 1)
+
+  def normalize: Array[Double] = centrality.map(c => c / (graph.maxNodeId - 1))
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -1,10 +1,21 @@
+/*
+ * Copyright 2014 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
 package com.twitter.cassovary.algorithms.centrality
 
 import com.twitter.cassovary.graph.DirectedGraph
 
-/**
- * Created by bmckown on 1/28/15.
- */
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+
   val centrality = new Array[Double](graph.maxNodeId + 1)
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -22,5 +22,5 @@ abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
    * Normalize the values
    * @return
    */
-  def normalize: Array[Double] = centrality.map(c => c / (graph.maxNodeId - 1))
+  def normalize: Array[Double] = centrality.map(c => c / (graph.nodeCount - 1))
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -13,17 +13,17 @@
  */
 package com.twitter.cassovary.algorithms.centrality
 
+import collection.immutable
 import com.twitter.cassovary.graph.{Node, DirectedGraph}
-import collection.immutable.IndexedSeq
 
 /**
  * Centrality algorithm abstraction that requires a directed graph {@code graph} be passed upon
  * instantiation.
- * @param graph A directed graph
+ * {@note The centrality values are stored in an internal array indexed by node's id. Hence, do not use for graphs with very large ids.}
  */
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
 
-  val centralityValues = new Array[Double](graph.maxNodeId + 1)
+  protected val centralityValues = new Array[Double](graph.maxNodeId + 1)
 
   /**
    * Get the specified node's centrality value
@@ -36,5 +36,5 @@ abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
    * Run the centrality calculation and update {@code centralityValues}
    * @return The centrality values of the graph
    */
-  def recalculate: IndexedSeq[Double]
+  def recalculate(): immutable.IndexedSeq[Double]
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Twitter, Inc.
+ * Copyright 2015 Twitter, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
  * file except in compliance with the License. You may obtain a copy of the License at
@@ -16,13 +16,15 @@ package com.twitter.cassovary.algorithms.centrality
 import com.twitter.cassovary.graph.{Node, DirectedGraph}
 
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
-  def centrality: Array[Double]
+
+  val centralityValues = new Array[Double](graph.maxNodeId + 1)
 
   /**
-   * Normalize the values
+   * Get the specified node's centrality value
+   * @param n Node
    * @return
    */
-  def normalize: Array[Double] = centrality.map(c => c / (graph.nodeCount - 1))
-
   def apply(n: Node): Double = centrality(n.id)
+
+  def centrality: Array[Double]
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -16,6 +16,5 @@ package com.twitter.cassovary.algorithms.centrality
 import com.twitter.cassovary.graph.DirectedGraph
 
 abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
-
   val centrality = new Array[Double](graph.maxNodeId + 1)
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
@@ -15,4 +15,7 @@ package com.twitter.cassovary.algorithms.centrality
 
 import com.twitter.cassovary.graph.Node
 
+/**
+ * Starting point for all centrality algorithms.
+ */
 trait Centrality extends (Node => Double)
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
@@ -1,10 +1,19 @@
+/*
+ * Copyright 2014 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
 package com.twitter.cassovary.algorithms.centrality
 
-import com.twitter.cassovary.graph.DirectedGraph
-
-/**
- * Created by bmckown on 1/28/15.
- */
 trait Centrality {
+
   def run: Array[Double]
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
@@ -13,7 +13,6 @@
  */
 package com.twitter.cassovary.algorithms.centrality
 
-trait Centrality {
+import com.twitter.cassovary.graph.Node
 
-  def run: Array[Double]
-}
+trait Centrality extends (Node => Double) {}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
@@ -1,0 +1,10 @@
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.DirectedGraph
+
+/**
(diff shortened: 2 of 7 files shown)


pankajgupta:
Will review in a day. @AnishShah @szymonm @plofgren -- Pls feel free to review as well:

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+trait Centrality {
+
+  def run: Array[Double]
```
szymonm: Maybe a bit more general:
```
def run: Function1[Int, Double]
```
This allows using Array, Map, lazy evaluation, ...

szymonm: And maybe apply as the method name? That would be consistent with other algorithms we have (ex. PageRank)

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,88 @@
...
+object OutDegreeCentrality {
+
+  /**
+   * This method determines the outdegree centrality of each node in the graph.  Currently, we only support
+   * calculating the ODC for each member of the graph and not for the graph on the whole.  Although this
+   * functionality would not be terribly difficult to add.
+   * @param graph A DirectedGraph instance
+   * @return
+   */
+  def apply(graph: DirectedGraph, normalize: Boolean = true): Array[Double] = {
```
szymonm: This method is almost same as InDegreeCentrality:apply

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,88 @@
...
+   * Run the outdegree centrality calculation for the graph.
+   * @return
+   */
+  def run: Array[Double] = {
+    graph.foreach { node =>
+      centrality(node.id) = node.outboundCount
+    }
+    centrality
+  }
+}
```
szymonm: pls insert newline at the end of each file.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2014 Twitter, Inc.
```
szymonm: pls use 2015 in the header

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
szymonm: we can allow also computing centrality for some of nodes by using:
```
def run: PartialFunction[Int, Double]
```

pankajgupta: I like Function1, but let's use PartialFunction1 when we have a case for it.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,26 @@
...
+import com.twitter.cassovary.graph.DirectedGraph
+
+abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+  val centrality = new Array[Double](graph.maxNodeId + 1)
+
+  /**
+   * Normalize the values
+   * @return
+   */
+  def normalize: Array[Double] = centrality.map(c => c / (graph.maxNodeId - 1))
```
pankajgupta: divide by nodeCount instead?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/Centrality.scala
pankajgupta: most concise is 
trait Centrality extends (Node => Double)

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.DirectedGraph
+
```
pankajgupta: delete extra line

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,88 @@
...
+
+sealed abstract class DegreeCentrality(graph: DirectedGraph) extends AbstractCentrality(graph)
+
```
pankajgupta: delete extra line

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
pankajgupta: yes, @bmckown please merge the two and reduce code redundancy

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,26 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.DirectedGraph
+
+abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+  val centrality = new Array[Double](graph.maxNodeId + 1)
```
pankajgupta: I'd change this to a 
def centrality: Array[Double]
and let the implementor implement (also see below)

pankajgupta: this will also need (if you do a trait below that extends Node=>Double)
def apply(n: Node) = centrality(n.id)

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,88 @@
...
+}
+
+private class InDegreeCentrality(graph: DirectedGraph) extends DegreeCentrality(graph) {
+
+  /**
+   * Run the indegree centrality calculation for the graph.
+   * @return
+   */
+  def run: Array[Double] = {
+    graph.foreach { node =>
```
pankajgupta: prefer the simpler
val centrality = graph map { node => node.inboundCount }

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
pankajgupta: e.g., may be use node.neighborCount(..)

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
bmckown: From my experience it is not common to divide by node count.  Most often, you will see normalization done by dividing each node's centrality by the maximum number of possible edges (maxNodeId - 1) to other nodes.  

As a reference, here is the documentation for in-degree centrality in python's networkx module:  http://networkx.lanl.gov/reference/generated/networkx.algorithms.centrality.in_degree_centrality.html#networkx.algorithms.centrality.in_degree_centrality

pankajgupta: thanks for the link. Seems like it is dividing by nodeCount-1, not maxNodeId-1

bmckown: Ahh I think I figured the two were one in the same.  Just for my clarity, does nodeCount include nodes in the graph that do not have edges connecting them to other nodes, or does it just include "active" nodes?

bmckown: It seems that nodeCount returns the number of "active" nodes in the graph (i.e. nodes that are connected to other nodes).  However, I would argue that dividing by (maxNodeId - 1) is the right call here.

Lets consider a graph of 15 nodes with a fully connected subgraph of 5 nodes, therefore there are ten nodes in the graph that are not connected.  The method nodeCount would tell me that there are 5 nodes in the graph (which is correct IF I only want to consider active connections), however, the method maxNodeId would tell me that there are 15 nodes in the graph.  So now the question becomes how do I "normalize" the in-degree and out-degree centrality?  I would argue that the "best" normalization would be to divide by the maximum number of *possible* connections per node (which would be 14) *not* the maximum number of possible connections among currently active nodes.  

Thoughts?

bmckown:
Updated with changes.  Waiting to hear from @pankajgupta to advise on normalization calculation.  Thanks for the comments gentlemen!  Very much appreciated.

pankajgupta:
I am not sure why you think nodeCount won't include a disconnected node. It
should. Did you run into this somewhere? If so, please provide example and
we will file it as a bug to be fixed.

The problem with using maxNodeId is that there might be nodes numbered say
1,4 and 1000. Then maxNodeId will be 1000.

bmckown:
Updated with correct normalization by nodeCount.  Thanks @pankajgupta!

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,30 @@
...
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+
+abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+
+  val centralityValues = new Array[Double](graph.maxNodeId + 1)
+
+  /**
+   * Get the specified node's centrality value
+   * @param n Node
+   * @return
```
pankajgupta: complete the comment

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,30 @@
...
+   */
+  def apply(n: Node): Double = centrality(n.id)
+
+  def centrality: Array[Double]
```
pankajgupta: come to think of it, your intent is to return an immutable collection here.
so this should be 
def centrality: collection.IndexedSeq[Double]

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,30 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+
+abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+
+  val centralityValues = new Array[Double](graph.maxNodeId + 1)
```
pankajgupta: make this protected?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  /**
+   * Get the specified node's centrality value
+   * @param n Node
+   * @return
+   */
+  def apply(n: Node): Double = centrality(n.id)
```
pankajgupta: probably accessing centralityValues is better?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,84 @@
...
+
+import com.twitter.cassovary.graph.DirectedGraph
+
+sealed abstract class DegreeCentrality(graph: DirectedGraph) extends AbstractCentrality(graph) {
+
+  /**
+   * Normalize the values
+   * @return
+   */
+  def normalize(normalize: Boolean = true): Array[Double] = {
```
pankajgupta: not recommended to use the same name for method and param

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,84 @@
...
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.DirectedGraph
+
+sealed abstract class DegreeCentrality(graph: DirectedGraph) extends AbstractCentrality(graph) {
+
+  /**
+   * Normalize the values
+   * @return
```
pankajgupta: comment what the return values are, or delete the @return line

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,84 @@
...
+   */
+  def normalize(normalize: Boolean = true): Array[Double] = {
+    if (normalize)
```
pankajgupta: this is a bit odd. If normalize param is false, why even call this method?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
pankajgupta: in fact if you don't use centralityValues, you will be calling the expensive calculation of centrality everytime you use apply()

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
pankajgupta: also I think it is better to rename this to centralityAllNodes and to have an explicit build()
i.e.,
...
def centralityAllNodes: collection.IndexedSeq[Double] = centralityAllValues
def recalculate() : collection.IndexedSeq[Double]

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Run the outdegree centrality calculation for the graph.
+   * @return
+   */
+  def centrality: Array[Double] = {
+    graph foreach { n => centralityValues(n.id) = n.outboundCount }
+    centralityValues
+  }
+}
```
pankajgupta: This is a lot of code duplication with only the 'dir' being different.
Consider just having one class:

class DegreeCentrality(graph: DirectedGraph, dir: GraphDir, normalize: Boolean = true) extends AbstractDegreeCentrality(graph) {
recalculate()
def recalculate() = {
val denom = if (normalize) graph.nodeCount - 1 else 1
graph map { node => centralityValues(node.id) = node.neighborCount(dir) / denom }
centralityAllNodes
}
}

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
szymonm: If you want an immutable collection you should use:
collection.immutable.IndexedSeq instead collection.IndexedSeq

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentralitySpec.scala
```diff
@@ -0,0 +1,38 @@
...
+ * Created by bmckown on 1/28/15.
+ */
+class DegreeCentralitySpec extends WordSpec with Matchers {
+
+  "Degree Centrality" should {
+
+    lazy val graph  = TestGraphs.g6WithEmptyNodes
+
+    lazy val unnormalizedInDegreeCentrality = Array(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 2.0, 3.0, 1.0, 3.0, 1.0)
+    lazy val unnormalizedOutDegreeCentrality = Array(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 3.0, 2.0, 1.0, 2.0, 1.0, 2.0)
```
szymonm: would be clearer if we have only proper values in the arrays (from index 11) and check only them.

Leading zeros may be a bit strange.

bmckown:
Updated and refactored.  Let me know what you think!

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,40 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+import collection.immutable.IndexedSeq
+
+/**
+ * Centrality algorithm abstraction that requires a directed graph {@code graph} be passed upon
+ * instantiation.
```
pankajgupta: Please add this to the comment:

@note The centrality values are stored in an internal array indexed by node's id. Hence, do not use for graphs with very large ids.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,40 @@
...
+ * @param graph A directed graph
```
pankajgupta: not very useful, delete

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+
+  val centralityValues = new Array[Double](graph.maxNodeId + 1)
```
pankajgupta: private

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,40 @@
...
+   * @param n Node
+   * @return Centrality value for the n-th node
+   */
+  def apply(n: Node): Double = centralityValues(n.id)
+
+  /**
+   * Run the centrality calculation and update {@code centralityValues}
+   * @return The centrality values of the graph
+   */
+  def recalculate: IndexedSeq[Double]
```
pankajgupta: should be () as per convention as the method changes internal state

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+import collection.immutable.IndexedSeq
```
pankajgupta: 1. imports need to be in alphabetical order.
2. convention is to just import collection.immutable

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
pankajgupta: immutable.indexedSeq

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph.DirectedGraph
+import com.twitter.cassovary.graph.GraphDir.GraphDir
+import collection.immutable.IndexedSeq
```
pankajgupta: alpha order

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,37 @@
...
+ * Calculate the degree centrality for a specific graph and direction.  If normalization
+ * is {@code true} then divide each node's degree centrality by the maximum number
+ * of possible connections within the graph.
+ * @param graph The graph object on which calculate the degree centrality
+ * @param dir The direction of the edges to include in the degree centrality calculation.
+ * @param normalize Pass true to normalize the values by the maximum number of possible connections.
+ */
+class DegreeCentrality(graph: DirectedGraph, dir: GraphDir, normalize: Boolean = true) extends AbstractCentrality(graph) {
+
+  recalculate
```
pankajgupta: ?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentrality.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  override def recalculate: IndexedSeq[Double] = {
```
pankajgupta: remove 'override'

cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
```diff
@@ -76,14 +76,22 @@ object TestGraphs {
...
+    NodeIdEdgesMaxId(14, Array(15)),
+    NodeIdEdgesMaxId(15, Array(10, 11)))
+
+  val nodeSeqWithEmpty = Seq(
+    NodeIdEdgesMaxId(0, Array.empty[Int]),
+    NodeIdEdgesMaxId(1, Array.empty[Int])
+  ) ++ nodeSeq
+
+
+  val nodeSeqIterator = () => nodeSeq.iterator
```
pankajgupta: bear in mind that this will slightly be changed once https://github.com/twitter/cassovary/pull/137 is merged. pls merge master in before your next PR.

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentralitySpec.scala
```diff
@@ -0,0 +1,52 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.graph._
+import org.scalatest.{Matchers, WordSpec}
+
+/**
```
pankajgupta: The convention is to not do 'created by' comments.

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentralitySpec.scala
```diff
@@ -0,0 +1,52 @@
...
+ * Created by bmckown on 1/28/15.
+ */
+class DegreeCentralitySpec extends WordSpec with Matchers {
+
+  "Degree Centrality" should {
+
+    lazy val graph = TestGraphs.g6WithEmptyNodes
```
pankajgupta: can be outside the should block and a member of the class

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/centrality/DegreeCentralitySpec.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    "return in-degree centrality values that are less than one for each node" in {
+      val idc = new DegreeCentrality(graph, GraphDir.InDir, true)
+      val centrality = idc.centralityValues
+      centrality(0)  shouldEqual 0.0
+      centrality(1)  shouldEqual 0.0
+      centrality(11) shouldEqual 0.2857142857142857
```
pankajgupta: checking for equality of double values is dangerous because of floating point precision. Prefer saying 
should be (0.285 +- 0.001)

bmckown:
Updated.

bmckown:
Updated.

pankajgupta:
Thanks for being patient with all the comments. This is ready to merge post the PR https://github.com/twitter/cassovary/pull/137 is merged (which is a breaking api change to ArrayBasedDirectedGraph so you will need to modify TestGraph slightly).

bmckown:
Thank you for being patient!  I appreciate all of the comments.  I will modify TestGraph once PR #137 is merged.  Thanks for the heads up.

pankajgupta:
https://github.com/twitter/cassovary/pull/137 is now in master

bmckown:
Updated.

Issue #140, opened 2015-01-29
Clarify support for Undirected Graphs


plofgren:
I have an application where I needed an undirected graph.  I actually didn't realize until today that "Mutual" in StoredGraphDir meant undirected and was starting to write my own UndirectedGraph wrapper class when I noticed it.  How should we make this more clear?  For example I could add a comment to StoredGraphDir along the lines of
`Mutual, // In a mutual graph, the outbound and inbound neighbors of each node are equal, and space is typically saved by only storing the outbound neighbors of each node`
and/or add to the documentation of DirectedGraph that it actually also supports undirected graphs.  Or should there be an UndirectedGraph wrapper in Cassovary?
@pankajgupta

pankajgupta:
Let's make this more explicit by adding a wrapper.

Alireza-Farahani:
I'm using version 7.1.0 and can't find anything explicitly for undirected graphs. What's the correct way of working with undirected graphs?
A simple graph like `1 2` when loaded by this code:
```scala
    val graphReader = ListOfEdgesGraphReader.forIntIds(
      dir, "toy_9nodes",
      graphDir = StoredGraphDir.Mutual 
    )
    val graph = graphReader.toArrayBasedDirectedGraph()
    println(graph)
 ```
prints
```txt
Node count: 2
Edge count: 2
Nodes:
NodeId => 1
InboundNodes[1] =>2|
OutboundNodes[1] =>2|

NodeId => 2
InboundNodes[0] =>
OutboundNodes[0] =>
```
which very confusing.

PR #141, opened 2015-01-30
Undirected Graph Support in Cassovary

plofgren:
Related to https://github.com/twitter/cassovary/issues/140
The idea of this wrapper is that each graph implementation shouldn't need custom code for `StoredGraphDir.Mutual`.  We can write a single wrapper which works with any graph implementation.
Here is an initial undirected graph class.  If this looks good, I can add a test and support for mutable undirected graphs to this pull request.
@pankajgupta

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
@@ -1,0 +1,26 @@
+/*
+* Copyright 2015 Twitter, Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+* file except in compliance with the License. You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software distributed
+* under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+* CONDITIONS OF ANY KIND, either express or implied. See the License for the
+* specific language governing permissions and limitations under the License.
+*/
+package com.twitter.cassovary.graph
+
+import com.twitter.cassovary.graph.node.UndirectedNode
+
+/**
+ * A graph where the outbound neighbors of each node equal the inbound neighbors.
+ *
+ */
+trait UndirectedGraph extends DirectedGraph {
+  val storedGraphDir = StoredGraphDir.Mutual
+  def getNodeById(id: Int): Option[UndirectedNode]
+  def iterator: Iterator[UndirectedNode]
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
@@ -1,0 +1,29 @@
+package com.twitter.cassovary.graph
+
+import com.twitter.cassovary.graph.StoredGraphDir.StoredGraphDir
+
+/*
+ * A graph where the outbound neighbors of each node equal the inbound neighbors.
+ * Given a DirectedGraph which stores the neighbors of each node (as outbound edges),
+ * this class wraps that graph in an undirected view.
+ */
+
+class UndirectedGraph(outboundGraph: DirectedGraph) extends DirectedGraph {
+  override def getNodeById(id: Int): Option[Node] = outboundGraph.getNodeById(id) map { new UndirectedNode(_)}
+
+  override def iterator: Iterator[Node] = outboundGraph.iterator map { new UndirectedNode(_)}
+
+  override def edgeCount: Long = outboundGraph.edgeCount
+
+  override val storedGraphDir: StoredGraphDir = outboundGraph.storedGraphDir
+
+  override def nodeCount: Int = outboundGraph.nodeCount
+}
+
+class UndirectedNode(outboundNode: Node) extends Node {
+  override def outboundNodes(): Seq[Int] = outboundNode.outboundNodes()
+
+  override def inboundNodes(): Seq[Int] = outboundNode.outboundNodes()
+
+  override val id: Int = outboundNode.id
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
@@ -1,5 +1,5 @@
 /*
-* Copyright 2014 Twitter, Inc.
+* Copyright 2015 Twitter, Inc.
 *
 * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
 * file except in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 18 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
```diff
@@ -0,0 +1,29 @@
+package com.twitter.cassovary.graph
```
pankajgupta: pls add the standard copyright header you can get from other files (except change 2014 to 2015)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import com.twitter.cassovary.graph.StoredGraphDir.StoredGraphDir
+
+/*
+ * A graph where the outbound neighbors of each node equal the inbound neighbors.
+ * Given a DirectedGraph which stores the neighbors of each node (as outbound edges),
+ * this class wraps that graph in an undirected view.
+ */
+
```
pankajgupta: delete extra newline

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
```diff
@@ -0,0 +1,29 @@
...
+class UndirectedGraph(outboundGraph: DirectedGraph) extends DirectedGraph {
```
pankajgupta: I'd make it a trait and not override methods that do not do anything.

trait UndirectedGraph extends DirectedGraph {
def storedGraphDir = Mutual
def getNodeById(id: Int): Option[UndirectedNode]
def iterator: Iterator[UndirectedNode]
}

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def iterator: Iterator[Node] = outboundGraph.iterator map { new UndirectedNode(_)}
+
+  override def edgeCount: Long = outboundGraph.edgeCount
+
+  override val storedGraphDir: StoredGraphDir = outboundGraph.storedGraphDir
+
+  override def nodeCount: Int = outboundGraph.nodeCount
+}
+
+class UndirectedNode(outboundNode: Node) extends Node {
```
pankajgupta: look at existing UnidirectionalNode.scala

UndirectedNode is a UnidirectionalNode with in and out bonded

so I'd put this inside UnidirectionalNode.scala

trait UndirectedNode extends UnidirectionalNode {
def inboundNodes() = outboundNodes()
}

and modify that scala file's objects' apply methods to return UndirectedNode when the direction is Mutual

while at it, please make UnidirectionalNode a trait instead of an abstract class.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.twitter.cassovary.graph.StoredGraphDir.StoredGraphDir
+
+/*
+ * A graph where the outbound neighbors of each node equal the inbound neighbors.
+ * Given a DirectedGraph which stores the neighbors of each node (as outbound edges),
+ * this class wraps that graph in an undirected view.
+ */
+
+class UndirectedGraph(outboundGraph: DirectedGraph) extends DirectedGraph {
+  override def getNodeById(id: Int): Option[Node] = outboundGraph.getNodeById(id) map { new UndirectedNode(_)}
```
szymonm: That's not about this PR, but some thoughts related.

Is there any gain from this wrapper (UndirectedGraph) if getNodeById (and other methods) returns Node (normal, not an uniderected one). We cannot use UndirectedNode properties (unless casting).

Alternatively, we can have:
```
trait Graph[NodeType <: Node] {
  getNodeById: Option[NodeType]
}
trait UndirectedGraph extends Graph[UndirectedNode] {
  getNodeById: Option[UndirectedNode]
}
```

pankajgupta:
@szymonm : I was also just thinking that, and agree that parametrizing Graph with NodeType will be better. Let's do that in a separate PR. See my comment above, we can do getNodeById to return UndirectedNode in UndirectedGraph regardless.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
plofgren: Done

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
plofgren: Done

cassovary-core/src/main/scala/com/twitter/cassovary/graph/UndirectedGraph.scala
plofgren: Why does it matter that we return an UndirectedNode, as opposed to a Node which happens to have the same outbound and inbound neighbors?  Relating to Szymon's comment, what properties do undirected nodes have that Nodes don't have?

pankajgupta: An UndirectedGraph is a collection of UndirectedNodes so an access operation on it should return UndirectedNode. e.g., accessing an element of List[Dog] should return an element of type Dog, not an Animal. This is independent of what Dog might have that an Animal does not have because codes are always fluid.

szymonm:
On the other hand, I used: https://github.com/scala-graph/scala-graph. 

They use:
```
trait GraphLike[N,
                E[X]  <: EdgeLikeIn[X],
                +This[X, Y[X]<:EdgeLikeIn[X]]
                      <: GraphLike[X,Y,This] with AnySet[Param[X,Y]] with Graph[X,Y]]
```
and IMO this type magic makes the library very hard to use...

pankajgupta:
:) Agreed that there's a balance. Parametrizing by NodeType is good I think.

plofgren:
(Sorry for the delay).  I changed UndirectedGraph to a trait, made UndirectedNode a sub-trait of UnidirectedNode, and made UndirectedGraph's methods return UndirectedNode.  I think that addresses all of your comments.  I also just rebased from master.

pankajgupta:
Ouch, somehow I missed your updated commits, Peter!
Also, looks your rebase has gone awry as the diff in this PR is showing other changes unrelated to your change. May I ask you to build a new patch or create a clean one on top of current master. Thanks.

pankajgupta:
Ping @plofgren

plofgren:
To be honest, I don't think I'll get back to this until at least after I finish my PhD in a couple months.  It doesn't really add functionality, so I propose we just drop it and optionally change the documentation as suggested in #140.

pankajgupta:
Ah, you have done a lot of work. Let's leave this open for now and hope you come back to it when you get time. Or someone else might pick it up too.

plofgren:
Sounds good.

Issue #139, opened 2015-01-29
Allow info to be stored per node in an array of nodes


pankajgupta:
Motivated by https://github.com/twitter/cassovary/pull/138 where a centrality score is being computed per node and it might be sometimes most efficient to keep information per node stored in an array.

Currently we have graph.tourist.InfoKeeper that keeps info in a mutable.Map which precludes storing infoPerNode in an array.

How about declaring a new trait that has get/put. and allowing infoPerNode to return a collection of this trait type. This can be either a mutable.Map (as now) or an Array with implicit conversions from the Map and Array to this trait? 

Comments @szymonm ?

PR #142, opened 2015-01-30
Fastutils wrapping benchmark

szymonm:


szymonm:
When I run the code wrapped version of fastutils collection performs 50% worse than not wrapped:
```
time of pure map ops: 1862
time of wrapped map ops: 2678
``` 
This is possibly due to boxing.

pankajgupta:
I verified on my computer as well, and it is indeed 50% slower on mine too...

pankajgupta:
Interestingly, I get the same results when I do
    val easyMap = (new Int2IntOpenHashMap()).asInstanceOf[java.util.Map[Int, Int]]

So it is not the scala wrapping, but just the boxing/unboxing at java level.

Any suggestions, @szymonm ? esp. in the light of #139 where also we want to rethink InfoKeeper.

szymonm:
Yes, looks like using Int2IntOpenHashMap as Map<Int, Int> makes the methods boxed.

Right, we have go back with InfoKeepers.

And hope for efficient specialization in Scala one day.

szymonm:
I'm afraid that this means whenever we use:
```
val nodes: Seq[Int] = new Array[Int](100)
```
then we have boxing...

szymonm:
```
scala> :paste
// Entering paste mode (ctrl-D to finish)

def time(b : => Unit) {
  val start = java.lang.System.currentTimeMillis()
  b
  print(s"Time: ${java.lang.System.currentTimeMillis() - start}")
}

// Exiting paste mode, now interpreting.

time: (b: => Unit)Unit

scala> import scala.util.Random
import scala.util.Random

scala> val arr = new Array[Int](1000000)
scala> (0 until 1000000).foreach{ i => arr(i) = i }

scala> var b = 0
b: Int = 0

scala> time {(0 until 1000000).foreach{ i => b = arr(i) * 7 }}
Time: 7
scala> val seq: Seq[Int] = arr
scala> time {(0 until 1000000).foreach{ i => b = seq(i) * 7 }}
Time: 19
```

szymonm:
@pankajgupta please have a look

pankajgupta:
Interesting indeed.

In general, in cassovary the desire has been to be first and foremost
efficient on storage (so native arrays) and then be as performant as
possible without getting too low level. This clearly shows that there are
opportunities to improve speed.

PR #148, opened 2015-02-04
Parameterize Graph by NodeType

AnishShah:
Is this Okay?

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/PageRank.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/PageRank.scala
@@ -13,7 +13,7 @@
  */
 package com.twitter.cassovary.algorithms
 
-import com.twitter.cassovary.graph.{DirectedGraph, GraphDir}
+import com.twitter.cassovary.graph.{DirectedGraph, GraphDir, Node}
 import com.twitter.cassovary.util.Progress
 import com.twitter.logging.Logger
 
@@ -42,7 +42,7 @@ object PageRank {
    * @param params PageRankParams from above
    * @return An array of doubles, with indices corresponding to node ids
    */
-  def apply(graph: DirectedGraph, params: PageRankParams): Array[Double] = {
+  def apply(graph: DirectedGraph[Node], params: PageRankParams): Array[Double] = {
     val pr = new PageRank(graph, params)
     pr.run
   }
@@ -54,13 +54,13 @@ object PageRank {
    * @param prArray An array of doubles, with indices corresponding to node ids
    * @return The updated array
    */
-  def iterate(graph: DirectedGraph, params: PageRankParams, prArray: Array[Double]) = {
+  def iterate(graph: DirectedGraph[Node], params: PageRankParams, prArray: Array[Double]) = {
     val pr = new PageRank(graph, params)
     pr.iterate(prArray: Array[Double])
   }
 }
 
-private class PageRank(graph: DirectedGraph, params: PageRankParams) {
+private class PageRank(graph: DirectedGraph[Node], params: PageRankParams) {
 
   private val log = Logger.get("PageRank")
 
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/TriangleCount.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/TriangleCount.scala
@@ -13,9 +13,8 @@
  */
 package com.twitter.cassovary.algorithms
 
-import com.twitter.cassovary.graph.DirectedGraph
+import com.twitter.cassovary.graph.{DirectedGraph, Node}
 import scala.util.Random
-import scala.collection.mutable.{ArrayBuffer, BitSet}
 import scala.collection.mutable
 
 case class TriangleCountParameters(edgeReservoirSize : Int, wedgeReservoirSize : Int)
@@ -28,7 +27,7 @@ case class TriangleCountParameters(edgeReservoirSize : Int, wedgeReservoirSize :
  * `graph` than (b, a) is also.
  */
 object TriangleCount {
-  def apply(graph: DirectedGraph, parameters: TriangleCountParameters, rng: Random = new Random): (Double, Double) = {
+  def apply(graph: DirectedGraph[Node], parameters: TriangleCountParameters, rng: Random = new Random): (Double, Double) = {
     new TriangleCount(graph, parameters, rng).apply()
   }
 }
@@ -76,7 +75,7 @@ class IntTuplesArray(noOfTuples : Int, tupleSize : Int) extends Iterable[Seq[Int
   }
 }
 
-class TriangleCount(graph : DirectedGraph, parameters : TriangleCountParameters, rng: Random) {
+class TriangleCount(graph : DirectedGraph[Node], parameters : TriangleCountParameters, rng: Random) {
 
   /**
    * We store an undirected edge as a list of ints `List(a, b)`, where `a < b`.
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
@@ -21,7 +21,7 @@ import com.twitter.cassovary.graph.{Node, DirectedGraph}
  * instantiation.
  * {@note The centrality values are stored in an internal array indexed by node's id. Hence, do not use for graphs with very large ids.}
  */
-abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+abstract class AbstractCentrality(graph: DirectedGraph[Node]) extends Centrality {
 
   protected val centralityValues = new Array[Double](graph.maxNodeId + 1)
 
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedGraph.scala
@@ -33,7 +33,7 @@ object StoredGraphDir extends Enumeration {
  * The entry point into a model of a directed graph.  Users typically query a known starting node
  * and then traverse the graph using methods on that {@code Node}.
  */
-trait DirectedGraph extends Graph with Iterable[Node] {
+trait DirectedGraph[+V <: Node] extends Graph[V] with Iterable[V] {
   /**
    * Returns the number of nodes in the graph.
    */
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Graph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Graph.scala
@@ -17,12 +17,12 @@ package com.twitter.cassovary.graph
  * The entry point into a model of a directed graph.  Users typically query a known starting node
  * and then traverse the graph using methods on that {@code Node}.
  */
-trait Graph {
+trait Graph[+V <: Node] {
   /**
    * Returns the node with the given {@code id} or else {@code None} if the given node does not
    * exist in this graph.
    */
-  def getNodeById(id: Int): Option[Node]
+  def getNodeById(id: Int): Option[V]
 
   def existsNodeId(id: Int) = getNodeById(id).isDefined
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/SharedArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/SharedArrayBasedDirectedGraph.scala
@@ -255,7 +255,7 @@ class SharedArrayBasedDirectedGraph private (nodeIdSet: Array[Byte], offsetTable
     lengthTable: Array[Int], sharedEdgeArray: Array[Array[Int]],
     reverseDirEdgeArray: Option[Array[Array[Int]]], maxId: Int, val nodeWithOutEdgesMaxId: Int,
     val nodeWithOutEdgesCount: Int, val nodeCount: Int, val edgeCount: Long,
-    val storedGraphDir: StoredGraphDir) extends DirectedGraph {
+    val storedGraphDir: StoredGraphDir) extends DirectedGraph[Node] {
 
   override lazy val maxNodeId = maxId
 
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
@@ -24,18 +24,18 @@ import scala.util.Random
 /**
  * A Traverser traverses the graph in a certain order of nodes.
  */
-trait Traverser extends Iterator[Node] { self =>
+trait Traverser[+V <: Node] extends Iterator[V] { self =>
   /**
    * Traverser only visits node Ids listed in some node's edges, and thus
    * we know that node must exist, safe to assume graph.getNodeById return non-None value
    */
-  protected def getExistingNodeById(graph: Graph, id: Int) = graph.getNodeById(id).get
+  protected def getExistingNodeById[W >: V <: Node](graph: Graph[W], id: Int): W = graph.getNodeById(id).get
 }
 
 /**
  * Bounds an iterator to go no more than a specified maximum number of steps
  */
-trait BoundedIterator[T] extends Iterator[T] {
+trait BoundedIterator[+T] extends Iterator[T] {
   /**
    * @return If option is defined, it is the maximal number of elements the iterator returns.
    *         If is `None`, the iterator is not bounded.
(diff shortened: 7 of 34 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -24,12 +24,12 @@ import scala.util.Random
 /**
  * A Traverser traverses the graph in a certain order of nodes.
  */
-trait Traverser extends Iterator[Node] { self =>
+trait Traverser[NodeType <: Node] extends Iterator[Node] { self =>
```
szymonm: why not Iterator[NodeType]?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Graph.scala
```diff
@@ -17,12 +17,12 @@ package com.twitter.cassovary.graph
  * The entry point into a model of a directed graph.  Users typically query a known starting node
  * and then traverse the graph using methods on that {@code Node}.
  */
-trait Graph {
+trait Graph[NodeType <: Node] {
```
szymonm: probably
```
trait Graph[+NodeType <: Node]
```
is what we want

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: also
```
trait Traverser[+NodeType <: Node]
```

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -396,7 +396,7 @@ trait QueueBasedTraverser extends Traverser {
 /**
  * A traverser that keeps track of first depth of visiting a given node.
  */
-trait DepthTracker extends QueueBasedTraverser {
+trait DepthTracker[NodeType <: Node] extends QueueBasedTraverser[NodeType] {
```
szymonm: I wondering if we cannot use just
```
trait DepthTracker extends QueueBasedTraverser[Node]
```
as DepthTracker itself does not use information about particular Node type.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedGraph.scala
```diff
@@ -33,7 +33,7 @@ object StoredGraphDir extends Enumeration {
  * The entry point into a model of a directed graph.  Users typically query a known starting node
  * and then traverse the graph using methods on that {@code Node}.
  */
-trait DirectedGraph extends Graph with Iterable[Node] {
+trait DirectedGraph extends Graph[Node] with Iterable[Node] {
```
pankajgupta: DirectedGraph should also be of type [+NodeType <: Node]

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Graph.scala
pankajgupta: prefer a shorter V instead of 'NodeType', so
trait Graph[+V <: Node]

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -71,10 +71,10 @@ trait BoundedIterator[T] extends Iterator[T] {
  * @param maxDepth if set, max depth of path
  * @param filterHomeNodeByNumEdges filter home node by number of edges
  */
-class RandomTraverser(graph: Graph, dir: GraphDir, homeNodeIds: Seq[Int],
+class RandomTraverser[NodeType <: Node](graph: Graph[NodeType], dir: GraphDir, homeNodeIds: Seq[Int],
```
pankajgupta: use +NodeType instead of NodeType everywhere.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: @szymonm while it technically does not have to be as you point out, there is a case for making it because the trait DepthTracker "is a" QueueBasedTraverser (because of 'extends' keyword) and it is a bit odd to not type parametrize it. e.g., here do we just want to have a QueueBasedTraverser of Node, and not of more specific NodeTypes? If so (and I am fine with it), your suggestion makes sense.

szymonm: QueueBasedTraverser should definitely be of NodeType.

pankajgupta: but DepthTracker only needs QueueBasedTraverser[Node] ? If so, okay to not
have type parametrization in DepthTracker.

AnishShah: so, just `QueueBasedTraverser[Node]` ? @szymonm ?

pankajgupta: Let's go with that for now.

szymonm: Thought about this. I think 
```
trait DepthTracker  extends QueueBasedTraverser[Node]
```
is ok. Thanks to covariance QueueBasedTraverser[NodeType] <: QueueBasedTraverser[Node] so we can extend it with DepthTracker.

AnishShah:
Please have a look. I'm not sure about `ArrayBasedDirectedGraph`, `SharedArrayBasedDirectedGraph` and also Benchmark files. But, I have checked everything and it works fine.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/PageRank.scala
```diff
@@ -42,7 +42,7 @@ object PageRank {
    * @param params PageRankParams from above
    * @return An array of doubles, with indices corresponding to node ids
    */
-  def apply(graph: DirectedGraph, params: PageRankParams): Array[Double] = {
+  def apply[V <: Node](graph: DirectedGraph[V], params: PageRankParams): Array[Double] = {
```
pankajgupta: No need to parametrize in this file. It uses only node.id that you can get by just Directedgraph[Node]

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/TriangleCount.scala
```diff
@@ -28,7 +27,7 @@ case class TriangleCountParameters(edgeReservoirSize : Int, wedgeReservoirSize :
  * `graph` than (b, a) is also.
  */
 object TriangleCount {
-  def apply(graph: DirectedGraph, parameters: TriangleCountParameters): (Double, Double) = {
+  def apply[V <: Node](graph: DirectedGraph[V], parameters: TriangleCountParameters): (Double, Double) = {
```
pankajgupta: Ditto, no need to parametrize.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/AbstractCentrality.scala
```diff
@@ -21,7 +21,7 @@ import com.twitter.cassovary.graph.{Node, DirectedGraph}
  * instantiation.
  * {@note The centrality values are stored in an internal array indexed by node's id. Hence, do not use for graphs with very large ids.}
  */
-abstract class AbstractCentrality(graph: DirectedGraph) extends Centrality {
+abstract class AbstractCentrality[+V <: Node](graph: DirectedGraph[V]) extends Centrality {
```
pankajgupta: Ditto, no need to parametrize.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SharedArrayBasedDirectedGraph.scala
```diff
@@ -255,7 +255,7 @@ class SharedArrayBasedDirectedGraph private (nodeIdSet: Array[Byte], offsetTable
     lengthTable: Array[Int], sharedEdgeArray: Array[Array[Int]],
     reverseDirEdgeArray: Option[Array[Array[Int]]], maxId: Int, val nodeWithOutEdgesMaxId: Int,
     val nodeWithOutEdgesCount: Int, val nodeCount: Int, val edgeCount: Long,
-    val storedGraphDir: StoredGraphDir) extends DirectedGraph {
+    val storedGraphDir: StoredGraphDir) extends DirectedGraph[Node] {
```
pankajgupta: for now, this is okay, but we will change it in a separate PR/issue.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -24,21 +24,21 @@ import scala.util.Random
...
  * A Traverser traverses the graph in a certain order of nodes.
  */
-trait Traverser extends Iterator[Node] { self =>
+trait Traverser[+V <: Node] extends Iterator[V] { self =>
   /**
    * Traverser only visits node Ids listed in some node's edges, and thus
    * we know that node must exist, safe to assume graph.getNodeById return non-None value
    */
-  protected def getExistingNodeById(graph: Graph, id: Int) = graph.getNodeById(id).get
+  protected def getExistingNodeById[V <: Node](graph: Graph[V], id: Int) = graph.getNodeById(id).get
```
pankajgupta: By defining a new [V <: Node] here you have shadowed the V in Traverser[V]

what you actually want here is:
getExistingNodeById[W >: V](graph: Graph[W]...)


see http://docs.scala-lang.org/tutorials/tour/variances.html

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -363,7 +363,7 @@ trait QueueBasedTraverser extends Traverser {
   /**
    * Performs action needed when visiting a node.
    */
-  protected def processNode(node : Node): Node = {
+  protected def processNode[V <: Node](node : V): V = {
```
pankajgupta: Same problem as above U/W discussion. See if you can simply have processNode(Node) here. Also, seems like the return value is not being used so far, so return Unit.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/IterativeLinkAnalyzer.scala
```diff
@@ -33,7 +33,7 @@ case class SuppliedNodeInfo(val node: Node, val initialIterationWeight: Double)
  * @param resetProbOnRight reset probability when iterating from right to left
  * @param numTopContributors The number of contributors (reasons) to be kept for each score
  */
-class IterativeLinkAnalyzer(graphUtils: GraphUtils, resetProbOnLeft: Double,
+class IterativeLinkAnalyzer[+V <: Node](graphUtils: GraphUtils[V], resetProbOnLeft: Double,
```
pankajgupta: leave as GraphUtils[Node]

cassovary-core/src/test/scala/com/twitter/cassovary/graph/GraphBehaviours.scala
```diff
@@ -19,7 +19,7 @@ import org.scalatest.{Matchers, WordSpec}
 trait GraphBehaviours extends Matchers {
```
pankajgupta: just make this trait of type [V <: Node] and simply use DirectedGraph[V] in all methods below without specifying [V<; Node] in every method

cassovary-core/src/test/scala/com/twitter/cassovary/graph/TraverserSpec.scala
```diff
@@ -248,7 +248,7 @@ class TraverserSpec extends WordSpec with MockitoSugar with Matchers {
     "yield all nodes in DFS order in expected order in unique id walk (inDir)" in {
       val dir = GraphDir.InDir
       val trav = new DepthFirstTraverser(graph, dir, Seq(10), Walk.Limits(None, None, Some(10L)))
-        with PathLengthTracker with DiscoveryAndFinishTimeTracker
+        with PathLengthTracker[Node] with DiscoveryAndFinishTimeTracker[Node]
```
pankajgupta: will be nice to add a test on any traverser on any subtype of Node to test the new parametrization.

pankajgupta:
We do need to move all DirectedGraph implementations to be parametrized by node type as well. But lets do it in a separate PR.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/TraverserSpec.scala
AnishShah: OK. Will do it.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
AnishShah: return value is used in `next()`

AnishShah: Also, can't keep `processNode(Node)` since return type of `next()` should be the subtype of `Node` .

Should I also parameterize `visitNode()` ?

cassovary-core/src/test/scala/com/twitter/cassovary/graph/TraverserSpec.scala
AnishShah: I need help with writing tests. Should I create `TestBipartiteGraph.scala` for `Bipartite Node` and `TestDynamicGraph.scala`  for `Dynamic Node` and then use it in `TraverserSpec.scala` ?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: In next(), instead of 
```processNode(getExistingNodeById(graph, nextId))```
perhaps do this:
```
...
val node = getExistingNodeById(graph, nextId)
processNode(node)
node
```

File: cassovary-core/src/test/scala/com/twitter/cassovary/graph/TraverserSpec.scala (hunk 14 above)
Comment (pankajgupta) [inline cassovary-core/src/test/scala/com/twitter/cassovary/graph/TraverserSpec.scala]: Just add a new bipartite graph to TestGraph.scala and use it in TraverserSpec. We are looking for one simple test to test the new functionality, and do not need all the methods / classes of Traverser tested again

**Comment by pankajgupta** [comment]
LGTM otherwise

File: cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala (hunk 10 above)
Comment (AnishShah) [inline cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala]: Not sure if it should be `W >: V` . It means, W should be either `V` or its super type. But, here V is only the sub types of Node, which breaks the type bound, I guess.

Comment (AnishShah) [inline cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala]: It should be `W <: Node`, I guess.

Comment (pankajgupta) [inline cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala]: No I think W >: V is right. A traverser of type V can do something with
graph of type V or supertype but can not of subtypes.

**Comment by AnishShah** [comment]
I'm busy this week. I won't be able to work on this until Friday. Sorry.

File: cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala (hunk 10 above)
Comment (AnishShah) [inline cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala]: I think, it is breaking the type bound.
I'm getting this error `type arguments [W] do not conform to trait Graph's type parameter bounds [+V <: com.twitter.cassovary.graph.Node]`

Here is some rough code
```
scala> trait Node
defined trait Node

scala> trait Graph[+V <: Node]
defined trait Graph

scala> trait Traverser[+V <: Node] {
     |   protected def someFunction[W >: V](graph: Graph[W]) = println("hi")
     | }
<console>:10: error: type arguments [W] do not conform to trait Graph's type parameter bounds [+V <: Node]
         protected def someFunction[W >: V](graph: Graph[W]) = println("hi")
 
```

pankajgupta: try [W >: V <: Node]

AnishShah: Yeah! That was working. Cool :)

AnishShah:
I have added the tests

1. I refactored the code and moved `BipartiteGraphSingleSide & DoubleSide` to TestGraphs and I'm using that in `BipartiteGraphSpec.scala` and `TraverserSpec.scala`. Is it OK?

2. While writing the test, I found out a bug. in [this line](https://github.com/AnishShah/cassovary/commit/776267fcac7645e4d3f0dc4fc1308f638771de24#diff-e7998641f7338821ff1199baf366994aL331) , `depth(node.id).get` fails for the first node since depth is not assigned to the first node. So, I changed it to getOrElse(-1). Am I correct? 

Please tell me if there are any other changes needed.

pankajgupta:
1. OK

2. Curious how you ran into the bug. there is   `enqueue(homeNodeIds, None)` to make sure homeNodeIds are assigned depth in DepthTracker.

AnishShah:
@pankajgupta So sorry, I misinterpreted the bug. It is actually because for Bipartite Graph LeftNode,
`enqueue(homeNodeIds, None)` saves depth for negative number and in `chooseNodesToEnqueue` it asks for depth of LeftNode but with positive number. 

I will need your help on this. Should I save negative id for LeftNode? What do you suggest?

pankajgupta:
No worries.
I think that might be a bug in BipartiteGraph (possibly where class LeftNode in BipartiteGraph.scala should have val id = -nodeId). I will suggest removing the Traverser tests for BipartiteGraph, closing this PR and tackling the traverser tests for BipartiteGraph by fixing this bug in a separate PR.

AnishShah:
Cool

pankajgupta:
Sorry I wasn't clear. Let's finish this PR first and merge to master without the BipartiteGraph traverser tests. Then add BipartiteGraph traverser tests in a separate PR.

AnishShah:
I have remove the tests, changed `getOrElse(-1)` to `get` but I have still kept the bipartiteGraph in TestGraphs.

Please tell me what changes are needed now?

pankajgupta:
@AnishShah please follow up with (1) the bipartite graph tests we removed from this PR, and (2) Wmove all DirectedGraph implementations to be parametrized by node type as well. Please make new issues for each one if you don't mind.

PR #150, opened 2015-02-11
Bug fix in mutable graph and edge reader

plofgren:
This pull request fixes a bug where ArrayBasedDynamicDirectedGraph couldn't be created from an adjacency list file. 

Currently the iterator in AdjacencyListGraphReader returns a reference to a single NodeIdEdgesMaxId which is mutated on the next iteration, so if the iterator is read into another collection like a sequence, the sequence will contain many references to the same object.  After this commit AdjacencyListGraphReader constructs a new NodeIdEdgesMaxId on each iteration, which may result in more garbage collection pressure, but won't result in bizzare bugs if users ever convert the iterator to a Seq.

Also, (this change)[https://github.com/twitter/cassovary/commit/8bfa6d96d33e83330dbee81a8675258d1a1a0a48] modified ArrayBasedDynamicDirectedGraph to that it reads reads the entire input into memory before reading any of it, but because I hadn't written a unit test for reading a file into a mutable graph, it was unclear that the `.view` was important for correctness.  I restored `.view` on the input iterator before flattening.   This change also fixes the bug, but I still think it also makes sense to pull my change to AdjacencyListGraphReader unless profiling tests have shown that reusing the same object on different iterations improves performance significantly.

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
@@ -37,7 +37,7 @@ import scala.collection.mutable
  * This case class holds a node's id, all its out edges, and the max
  * id of itself and ids of nodes in its out edges
  */
-case class NodeIdEdgesMaxId(val id: Int, val edges: Array[Int], val maxId: Int)
+case class NodeIdEdgesMaxId(id: Int, edges: Array[Int], maxId: Int)
 
 object NodeIdEdgesMaxId {
   def apply(id: Int, edges: Array[Int]) =
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
@@ -37,7 +37,7 @@ import scala.collection.mutable
  * This case class holds a node's id, all its out edges, and the max
  * id of itself and ids of nodes in its out edges
  */
-case class NodeIdEdgesMaxId(var id: Int, var edges: Array[Int], var maxId: Int)
+case class NodeIdEdgesMaxId(val id: Int, val edges: Array[Int], val maxId: Int)
 
 object NodeIdEdgesMaxId {
   def apply(id: Int, edges: Array[Int]) =
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
@@ -46,7 +46,8 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
 
   def this(iterableSeq: Seq[Iterable[NodeIdEdgesMaxId]],
            storedGraphDir: StoredGraphDir) {
-    this(iterableSeq.flatten , storedGraphDir)
+    // Use view to avoid eagerly reading all edges into memory
+    this(iterableSeq.view.flatten.toIterable, storedGraphDir)
   }
 
   /* Returns an option which is non-empty if outbound list for id  is non-null. */
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReader.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReader.scala
@@ -82,7 +82,6 @@ class AdjacencyListGraphReader[T] (
       private val src = Source.fromFile(filename)
       private val lines = src.getLines()
         .map{x => {lastLineParsed += 1; x}}
-      private val holder = NodeIdEdgesMaxId(-1, null, -1)
 
       override def hasNext: Boolean = {
         val isNotLastLine = lines.hasNext
@@ -110,10 +109,7 @@ class AdjacencyListGraphReader[T] (
             i += 1
           }
 
-          holder.id = internalNodeId
-          holder.edges = outEdgesArr
-          holder.maxId = newMaxId
-          holder
+          NodeIdEdgesMaxId(internalNodeId, outEdgesArr, newMaxId)
         } catch {
           case NonFatal(exc) =>
             throw new IOException("Parsing failed near line: %d in %s"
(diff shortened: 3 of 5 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
```diff
@@ -46,7 +46,8 @@ class ArrayBasedDynamicDirectedGraph(val storedGraphDir: StoredGraphDir)
 
   def this(iterableSeq: Seq[Iterable[NodeIdEdgesMaxId]],
            storedGraphDir: StoredGraphDir) {
-    this(iterableSeq.flatten , storedGraphDir)
+    // Use view to avoid eagerly reading all edges into memory
+    this(iterableSeq.view.flatten.toIterable, storedGraphDir)
```
szymonm: do you need toIterable at the end then?

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReader.scala
```diff
@@ -111,10 +110,7 @@ class AdjacencyListGraphReader[T] (
             i += 1
           }
 
-          holder.id = internalNodeId
-          holder.edges = outEdgesArr
-          holder.maxId = newMaxId
-          holder
+          NodeIdEdgesMaxId(internalNodeId, outEdgesArr, newMaxId)
```
szymonm: For the case of immutable graph we would prefer to avoid creating the object here, don't we @pankajgupta ?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDynamicDirectedGraph.scala
plofgren: Yes, I don't know why, but otherwise it doesn't compile.  The type without the toIterable seems to be
`scala.collection.SeqView[Iterable[com.twitter.cassovary.graph.NodeIdEdgesMaxId],Seq[Iterable[com.twitter.cassovary.graph.NodeIdEdgesMaxId]]]#Transformed[com.twitter.cassovary.graph.NodeIdEdgesMaxId]`
It's quite a type :)

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReader.scala
plofgren: We might prefer not creating the object (I'll let Pankaj answer that) but I don't think it should matter whether the graph is immutable or not. The question is whether the NodeIdEdgesMaxId returned by each iteration should be immutable or not.

pankajgupta:
Having those vars in NodeIdEdgesMaxId was indeed a ticking time bomb and I'd prefer to have a new object being created every time. Please change those vars in class NodeIdEdgesMaxID() to vals while you are doing this. This will also uncover any other current unintended uses of those vars. I think there is one in ListOfEdgesGraphReader with the fix same as the one you have in AdjacencyListGraphReader, so please do that as well if you don't mind.

Having said that, it will still be good to know if there is a performance penalty to this change. Please report results from PerformanceBenchmark in cassovary-benchmarks with/without this change, to not run blind.

Thanks for the bug fix!

pankajgupta:
ping @plofgren

plofgren:
Sorry for the delay (I'm trying to finish two PhD projects, TA a class, and work part-time for a start-up).  Making the change to ListOfEdgesGraphReader was easy, but I haven't found time to learn to use the benchmark.  Do you or @szymonm happen to know the exact command to run the benchmarks?

pankajgupta:
Just added cassovary-benchmarks/BenchmarksREADME.md -- please feel free to
modify and look at the individual sources of benchmarks as well.

plofgren:
I just rebased off master and ran the benchmarks.  Thanks for the readme.  I used the Snap Youtube graph with 3 million edges because it ran in a reasonable amount of time (3 minutes).  I ran the benchmark on an ec2 m3.large instance (2 cpus, 7.5 GB Ram) which was not running any other jobs.  Executing
```runMain com.twitter.cassovary.PerformanceBenchmark -url=http://snap.stanford.edu/data/bigdata/communities/com-youtube.ungraph.txt.gz```
on twitter/master took 3 minutes and 20 seconds, and when I repeated it it again took 3 minutes and 20 seconds.  The same command on branch plofgren-mutable-graph-reader also took 3 minutes and 20 seconds on both trials.  I'm actually surprised how little variance there is in the running time (the largest difference across all 4 runs was only 100 ms). 

I'm seeing a message that the build failed, but I don't think it's related to this change (there is also a message on the main cassovary page saying the build is failing).  @pankajgupta let me know if you think the failed checks are related to this change.

Let me know if there's anything else I should do to get this merged into master.

**Comment by pankajgupta** [comment]
I'm on vacation this week but will check it out right after.

File: cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala [hunk 3]
```diff
@@ -37,7 +37,7 @@ import scala.collection.mutable
  * This case class holds a node's id, all its out edges, and the max
  * id of itself and ids of nodes in its out edges
  */
-case class NodeIdEdgesMaxId(var id: Int, var edges: Array[Int], var maxId: Int)
+case class NodeIdEdgesMaxId(val id: Int, val edges: Array[Int], val maxId: Int)
```
bmckown: I believe the val modifier is redundant here since you have a case class

pankajgupta:
Regarding the failed Travis CI build, there seems to be an issue irrespective of your change. Just filed issue #180, will hopefully track it down in the next few days

cassovary-core/src/test/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReaderSpec.scala
```diff
@@ -42,6 +42,11 @@ class AdjacencyListGraphReaderSpec extends WordSpec with Matchers with GraphBeha
     val graph = reader.toSharedArrayBasedDirectedGraph()
   }
 
+  trait ArrayBasedDynamicGraphWithoutRenumberer {
+    val graph = AdjacencyListGraphReader.forIntIds(directory, "toy_6nodes_adj",
+      Executors.newFixedThreadPool(2)).toArrayBasedDynamicDirectedGraph()
```
pankajgupta: not sure how this compiles as the third argument "Executors..." was deprecated in master. Please try running test on your latest commit.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
plofgren: I agree it's redundant.  Do you think it's better style to omit val?

plofgren:
Good point.  I didn't double-check `sbt test` after rebasing.  I just removed `Executors...` and now `sbt test` is passing.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala
pankajgupta: yes, please omit val.

pankajgupta:
Looks good, will merge after the CI issue is resolved #180

PR #152, opened 2015-02-12
Adding ability to add labels to ids (such as nodes).

pankajgupta:
We sometimes want to be able to store associated information with nodes and edges. This PR adds this facility and also includes a basic file reader for int node labels as an illustration. The file reading capability is not meant to be general (i.e., for all label types) at this point. Will also add an example to cassovary-examples next.

CC @szymonm @plofgren


cassovary-core/src/main/scala/com/twitter/cassovary/util/io/IoUtils.scala
```diff
@@ -0,0 +1,29 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.util.io
+
+import java.io.File
+
+object IoUtils {
+
+  def readFiles(directory: String, prefixFileNames: String): Array[String] = {
```
szymonm: Maybe readFileNames?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Labels.scala
```diff
@@ -0,0 +1,35 @@
...
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.graph.labels
+
+import scala.collection.mutable
+import scala.reflect.runtime.universe._
+
+class Labels[ID] {
```
szymonm: could ID be specialized for Int?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Labels.scala
```diff
@@ -0,0 +1,35 @@
...
+  private val underlying = mutable.HashMap.empty[String, Label[ID, _]]
+
+  def get[L : TypeTag](key: String): Option[Label[ID, L]] = {
+    underlying.get(key).flatMap { v =>
+      if (typeTag[L] == v.tag) Some(v.asInstanceOf[Label[ID, L]]) else None
```
szymonm: Should't we return None when v.asInstanceOf fails? Would be good to specify this in the scaladoc.

pankajgupta:
oh, failing travis ci build -- looks like I need to add a dependency in build file. Will do

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/IoUtils.scala
pankajgupta: sure

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Labels.scala
pankajgupta: will do

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Labels.scala
pankajgupta: actually it will never fail because it is type safe. That's the reason for the typeTag check.

szymonm: right, sorry, it was a bit late.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Label.scala
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// the simplest of labels backed by a mutable Hashmap
+class MapBasedLabel[ID, L](val name: String)(implicit val tag: TypeTag[L])
+    extends mutable.HashMap[ID, L] with Label[ID, L]
+
+// mix in this trait to make a label that is defined only for some ids
+trait PartialLabel[ID, L] extends Label[ID, L] {
+  def underlying: mutable.Set[ID]
+  private def isThere(id: ID): Boolean = underlying.contains(id)
```
szymonm: PartialLabel sounds similar to PartialFunction maybe isThere should be isDefinedAt then?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Graph.scala
```diff
@@ -25,4 +27,6 @@ trait Graph {
   def getNodeById(id: Int): Option[Node]
 
   def existsNodeId(id: Int) = getNodeById(id).isDefined
+
+  val nodeLabels = new Labels[Int] //empty to start with
```
szymonm: I'm wondering if it wouldn't be better to add labels as a trait and keep the possibility to have a graph with no labels.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/ArrayBasedLabel.scala
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * A subtype of Label that uses an underlying array to map an integer id to the value
+ * associated with that id. Useful when `maxId` is not too big.
+ * @see An array based label must be defined for all ids. If desired for only some ids,
+ *      mix it in with `PartialLabel`
+ * @param name  name of the label
+ * @param maxId maximum value of the id
+ * @tparam L    type of the label value
+ */
+class ArrayBasedLabel[L : ClassTag](val name: String, val maxId: Int)(implicit val tag: TypeTag[L])
```
szymonm: here Int-specialization also can make a difference

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/ArrayBasedLabel.scala
```diff
@@ -0,0 +1,73 @@
...
+    extends Label[Int, L] with HasMaxId {
+  private val array = new Array[L](maxId + 1)
+
+  def get(id: Int): Option[L] = Some(array(id))
```
szymonm: however using option here will cause boxing as Option is not specialized

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -27,6 +27,11 @@ trait Node {
   val id: Int
 
   /**
+   * The label of a node accessed by name. Label can be anything.
+   */
+  def label(name: String): Option[_] = None
```
szymonm: This method is never used so far, isn't it?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Label.scala
```diff
@@ -0,0 +1,75 @@
...
+  val name: String
+  val tag: TypeTag[L]
+}
+
+// the simplest of labels backed by a mutable Hashmap
+class MapBasedLabel[ID, L](val name: String)(implicit val tag: TypeTag[L])
+    extends mutable.HashMap[ID, L] with Label[ID, L]
+
+// mix in this trait to make a label that is defined only for some ids
+trait PartialLabel[ID, L] extends Label[ID, L] {
```
szymonm: It a label is a map than I understand that it is defined only for some ids. Why to have PartialLabel then?

szymonm: Note that mutable.Map is a PartialFunction

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Graph.scala
pankajgupta: Moving it to DirectedGraph. This allows Graph to be purer without any vals. Not much overhead to have an empty class included all the time for a collection like graph.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/Label.scala
pankajgupta: Correct that label is defined as partial in the first place. PartialLabel is only a sample implementation of label, useful for creating a Flaglabel for instance.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/labels/ArrayBasedLabel.scala
pankajgupta: Actually, choosing not to specialize at all rightaway. At least in terms of storage, native arrays will be used. Will need to do performance testing before doing specialization.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
pankajgupta: Correct, deleting it.

pankajgupta:
@szymonm looks okay?

szymonm:
LGTM

PR #154, opened 2015-02-16
Resolves 144

szymonm:
TriangleCountSpec results are now deterministic.

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
@@ -14,8 +14,10 @@
 package com.twitter.cassovary.graph
 
 import com.twitter.cassovary.graph.StoredGraphDir._
-import com.twitter.cassovary.util.{Sampling, BinomialDistribution}
+import com.twitter.cassovary.util.{BoundedFuturePool, Sampling, BinomialDistribution}
 import java.util.concurrent.ConcurrentLinkedQueue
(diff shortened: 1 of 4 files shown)


cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/TriangleCountSpec.scala
```diff
@@ -23,6 +25,10 @@ class TriangleCountSpec extends WordSpec with Matchers {
     (sums._1 / repetitions, sums._2 / repetitions)
   }
 
+  val random = new Random()
+
+  def rng(): Random = {random.setSeed(1); random}
```
pankajgupta: prefer not reusing 'val random' across tests. so create new object everytime in rng()

cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
```diff
@@ -14,8 +14,10 @@
 package com.twitter.cassovary.graph
 
 import com.twitter.cassovary.graph.StoredGraphDir._
-import com.twitter.cassovary.util.{Sampling, BinomialDistribution}
+import com.twitter.cassovary.util.{BoundedFuturePool, Sampling, BinomialDistribution}
 import java.util.concurrent.ConcurrentLinkedQueue
+import com.twitter.util.{Await, Future, FuturePool}
```
pankajgupta: alpha order

pankajgupta:
LGTM otherwise

PR #124, opened 2015-01-22
Migration from Ostrich to Metrics: Fix typos

AnishShah:
I'm really sorry for the typos.

PR #155, opened 2015-02-17
hasp map usage corrected

nextLane:
Trying to understand how Cassovary works, found this compile time error while trying examples. Fixed it.

### DIFF
--- a/cassovary-examples/src/main/java/RandomWalkJava.java
+++ b/cassovary-examples/src/main/java/RandomWalkJava.java
@@ -91,7 +91,7 @@ public class RandomWalkJava {
   @SuppressWarnings("unchecked")
   private static Map<Integer, Integer> scalaIntsMapToJavaMap(scala.collection.Map<Object, Object> map) {
     Map<Object, Object> javaMap = JavaConverters$.MODULE$.mapAsJavaMapConverter(map).asJava();
-    HashMap<Integer, Integer> javaIntsMap = new HashMap<>();
+    HashMap<Integer, Integer> javaIntsMap = new HashMap<Integer, Integer>();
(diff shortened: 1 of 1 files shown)


cassovary-examples/src/main/java/RandomWalkJava.java
```diff
@@ -91,7 +91,7 @@ public static void main(String[] args) {
   @SuppressWarnings("unchecked")
   private static Map<Integer, Integer> scalaIntsMapToJavaMap(scala.collection.Map<Object, Object> map) {
     Map<Object, Object> javaMap = JavaConverters$.MODULE$.mapAsJavaMapConverter(map).asJava();
-    HashMap<Integer, Integer> javaIntsMap = new HashMap<>();
+    HashMap<Integer, Integer> javaIntsMap = new HashMap<Integer, Integer>();
```
bmckown: Which version of java are you using?  To my knowledge that should compile for java 7 and up.  

http://docs.oracle.com/javase/7/docs/technotes/guides/language/type-inference-generic-instance-creation.html

AnishShah: Yep. HashMap<>() compiles for Java 7 and up.
And after pull #124 , cassovary works for Java 7 and up.

cc @pankajgupta

szymonm: yes, diamond notation was introduced in 7

nextLane: Yes, my bad, I was using JAVA 6, updated to 8. Works well now.

pankajgupta:
Updated README to clarify this.

PR #164, opened 2015-02-20
Tests for Bipartite Graph Traverser

AnishShah:


cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -62,8 +61,8 @@ class LeftNode(nodeId: Int, val inboundNodes: Seq[Int],
  * all of its in and out edges point to nodes in the LHS, and thus all edge ids have values
  * of the real id * -1
  * @param id the id of the node (unique on the RHS)
- * @param in original (positive) ids of the nodes on the LHS pointed by in-coming edges
- * @param out original (positive) ids of the nodes on the LHS pointed by out-going edges
+ * @param in original (negative) ids of the nodes on the LHS pointed by in-coming edges
```
pankajgupta: this is a bit confusing, but "original" refers to positive ids only. Negative ids are meant to be purely an internal construct.

pankajgupta: oh, I take it back. The negative ids are leaked outside as your tests show. I think it would be best to do away completely with this negative/positive translation inside the nodes, but that again is a separate PR...

PR #166, opened 2015-02-23
Add betweenness centrality and shortest path

bmckown:
New clean pull request for betweenness centrality that is up to date with master (specifically the parameterization by node type).  

This is still a WIP.  Please leave some comments on how to make this better as betweenness centrality is a rather complicated thing to calculate.

In plain English, we want to find all of the shortest paths from node i to node j.  We then collect all of the middle nodes between node i and node j not including i or j.  We then sum over those nodes dividing by the total number of shortest paths.  We do this for all nodes i, j that are contained in the graph.  

The most complicated part of this is calculating all of the shortest paths simultaneously in a timely fashion.  The code may be a little opaque due to this.  We could sacrifice efficiently for transparency if we determine that to be the best route.

@pankajgupta @szymonm @AnishShah please feel free to provide feedback.  Thanks!

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2015 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.algorithms.shortestpath.SingleSourceShortestPath
+import com.twitter.cassovary.graph._
+
+class BetweennessCentrality(graph: DirectedGraph[Node], normalize: Boolean = true) extends AbstractCentrality(graph) {
+
+  private def combineMap(a: Map[Int, Double], b: Map[Int, Double]): Map[Int, Double] = {
+    (a.keys ++ b.keys).map { k => k -> (a.getOrElse(k, 0.0) + b.getOrElse(k, 0.0)) }.toMap
+  }
+
+  def _recalc(): Unit = {
+    val values = graph.foldLeft(Map.empty[Int, Double]){ (partialValues, n) =>
+      val allPaths = SingleSourceShortestPath(graph, n.id).allShortestPaths
+      val currentValues = allPaths.values.foldLeft(Map.empty[Int, Double]){ (acc, pc) =>
+        val middleNodes = pc.flatMap { p => p.slice(1, p.size - 1) }
+          .groupBy(k => k)
+          .map { k => k._1 -> (k._2.size.toDouble / pc.length) }.toMap
+        combineMap(middleNodes, acc)
+      }
+      combineMap(partialValues, currentValues)
+    }
+    val nc = graph.nodeCount
+    val scale = if (normalize && nc >= 2) 1.0 / ((nc - 1) * (nc - 2)) else 1.0
+    values.foreach { case (i, v) => centralityValues(i) = v * scale }
+  }
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
@@ -19,7 +19,7 @@ import com.twitter.cassovary.graph._
 class BetweennessCentrality(graph: DirectedGraph[Node], normalize: Boolean = true) extends AbstractCentrality(graph) {
 
   private def combineMap(a: Map[Int, Double], b: Map[Int, Double]): Map[Int, Double] = {
-    (a.keys ++ b.keys).map { k => k -> (a.getOrElse(k, 0.0) + b.getOrElse(k, 0.0)) }.toMap
+    (a.keySet ++ b.keySet).map { k => k -> (a.getOrElse(k, 0.0) + b.getOrElse(k, 0.0)) }.toMap
   }
 
   def _recalc(): Unit = {
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
@@ -26,7 +26,7 @@ class BetweennessCentrality(graph: DirectedGraph[Node], normalize: Boolean = tru
     val values = graph.foldLeft(Map.empty[Int, Double]){ (partialValues, n) =>
       val allPaths = new SingleSourceShortestPath(graph, n.id).allShortestPaths
       val currentValues = allPaths.values.foldLeft(Map.empty[Int, Double]){ (acc, pc) =>
-        val middleNodes = pc.flatMap { p => p.slice(1, p.size - 1) }
+        val middleNodes = pc.flatMap { p => p.slice(1, p.size) }
           .groupBy(k => k)
           .map { k => k._1 -> (k._2.size.toDouble / pc.length) }.toMap
         combineMap(middleNodes, acc)
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
@@ -24,7 +24,7 @@ class BetweennessCentrality(graph: DirectedGraph[Node], normalize: Boolean = tru
 
   def _recalc(): Unit = {
     val values = graph.foldLeft(Map.empty[Int, Double]){ (partialValues, n) =>
-      val allPaths = SingleSourceShortestPath(graph, n.id).allShortestPaths
+      val allPaths = new SingleSourceShortestPath(graph, n.id).allShortestPaths
       val currentValues = allPaths.values.foldLeft(Map.empty[Int, Double]){ (acc, pc) =>
         val middleNodes = pc.flatMap { p => p.slice(1, p.size - 1) }
           .groupBy(k => k)
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
@@ -1,15 +1,40 @@
+/*
+ * Copyright 2015 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
 package com.twitter.cassovary.algorithms.shortestpath
 
 import com.twitter.cassovary.graph.{Node, DirectedGraph}
 
+/**
+ * A base trait for all ShortestPath algorithm implementations
+ */
 trait ShortestPath {
   type Path  = Seq[Int]
-  type Paths = Seq[Seq[Int]]
-  type Stack = List[Seq[Int]]
+  type Paths = Seq[Path]
+  type Stack = List[Path]
 
   def graph: DirectedGraph[Node]
 
-  def shortestPaths(n: Int): Paths
+  /**
+   * Calculate the shortest path to a given target
+   * @param target The integer ID of the target node
+   * @return A sequence of paths.  If no shortest path exists, the list will be empty
+   */
+  def shortestPaths(target: Int): Paths
 
-  def allShortestPaths: Map[Int, Paths] = graph.toList.map { n => n.id -> shortestPaths(n.id) }.toMap
+  /**
+   * Calculate all shortest paths for all target nodes in the graph
+   * @return A map of integer node id to paths
+   */
+  def allShortestPaths: Map[Int, Paths] = graph.map { n => n.id -> shortestPaths(n.id) }.toMap
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
@@ -20,7 +20,6 @@ import com.twitter.cassovary.graph.{Node, DirectedGraph}
  */
 trait ShortestPath {
   type Path  = Seq[Int]
-  type Stack = List[Path]
 
   val graph: DirectedGraph[Node]
 
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
@@ -20,21 +20,26 @@ import com.twitter.cassovary.graph.{Node, DirectedGraph}
  */
 trait ShortestPath {
   type Path  = Seq[Int]
-  type Paths = Seq[Path]
   type Stack = List[Path]
 
-  def graph: DirectedGraph[Node]
+  val graph: DirectedGraph[Node]
 
   /**
-   * Calculate the shortest path to a given target
+   * Source node for all shortest paths calculated by `shortestPaths`
+   */
+  val source: Int
+
+  /**
+   * Calculate the shortest paths to `target` from a given source node.  The source
+   * node must be provided as an instance variable of all implementing subclasses.
    * @param target The integer ID of the target node
    * @return A sequence of paths.  If no shortest path exists, the list will be empty
    */
-  def shortestPaths(target: Int): Paths
+  def shortestPaths(target: Int): Seq[Path]
 
   /**
    * Calculate all shortest paths for all target nodes in the graph
    * @return A map of integer node id to paths
    */
-  def allShortestPaths: Map[Int, Paths] = graph.map { n => n.id -> shortestPaths(n.id) }.toMap
+  def allShortestPaths: Map[Int, Seq[Path]] = graph.map { n => n.id -> shortestPaths(n.id) }.toMap
 }
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
@@ -1,0 +1,15 @@
+package com.twitter.cassovary.algorithms.shortestpath
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+
+trait ShortestPath {
+  type Path  = Seq[Int]
+  type Paths = Seq[Seq[Int]]
+  type Stack = List[Seq[Int]]
+
+  def graph: DirectedGraph[Node]
+
+  def shortestPaths(n: Int): Paths
+
+  def allShortestPaths: Map[Int, Paths] = graph.toList.map { n => n.id -> shortestPaths(n.id) }.toMap
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
@@ -1,0 +1,44 @@
+package com.twitter.cassovary.algorithms.shortestpath
+
+import com.twitter.cassovary.graph.{Node, GraphDir, BreadthFirstTraverser, DirectedGraph}
+
+case class SingleSourceShortestPath(graph: DirectedGraph[Node], source: Int) extends ShortestPath {
+
+  val bfs = new BreadthFirstTraverser(graph, GraphDir.OutDir, Seq(source))
+  private val (nodes, depths)  = (bfs.toList, bfs.allDepths)
+  private val previous = nodes.map { n =>
+    n.id -> n.inboundNodes().filter { in => depths(in) == depths(n.id) - 1}
+  }.toMap
+
+  private def walkOnGraph(currStack: Stack, currTop: Int): (Stack, Int, Path) = {
+    val node  = currStack(currTop)(0)
+    val index = currStack(currTop)(1)
+
+    val path = if (node == source)
+      currStack.take(currTop + 1).reverse.map { k => k(0) }.toSeq
+    else
+      Seq.empty[Int]
+
+    val (stack, top) = if (previous(node).length > index) {
+      val newStack = if (currTop + 1 == currStack.length)
+        currStack :+ Seq(previous(node)(index), 0)
+      else
+        currStack.zipWithIndex.map { case (seq, in) => if (in == currTop + 1) Seq(previous(node)(index), 0) else seq}
+      (newStack, currTop + 1)
+    }
+    else {
+      val newStack = currStack.zipWithIndex.map { case (seq, in) => if (in == currTop - 1) Seq(seq(0), seq(1) + 1) else seq}
+      (newStack, currTop - 1)
+    }
+    (stack, top, path)
+  }
+
+  def shortestPaths(target: Int): Paths = {
+    lazy val stream: Stream[(Stack, Int, Path)] =
+      (List(Seq(target, 0)), 0, Seq.empty[Int]) #:: stream.map { case (s, t, p) => walkOnGraph(s,t) }
+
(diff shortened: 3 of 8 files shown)


bmckown:
Thoughts @pankajgupta ?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,15 @@
...
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+
+trait ShortestPath {
+  type Path  = Seq[Int]
+  type Paths = Seq[Seq[Int]]
+  type Stack = List[Seq[Int]]
+
+  def graph: DirectedGraph[Node]
+
+  def shortestPaths(n: Int): Paths
```
pankajgupta: pls add comments. Is 'n' source or target?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,15 @@
+package com.twitter.cassovary.algorithms.shortestpath
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+
+trait ShortestPath {
+  type Path  = Seq[Int]
+  type Paths = Seq[Seq[Int]]
```
pankajgupta: should be able to reuse Path here. i.e.,
type Paths = Seq[Path]

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,15 @@
...
+  type Stack = List[Seq[Int]]
+
+  def graph: DirectedGraph[Node]
+
+  def shortestPaths(n: Int): Paths
+
+  def allShortestPaths: Map[Int, Paths] = graph.toList.map { n => n.id -> shortestPaths(n.id) }.toMap
```
pankajgupta: toList is not required

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
```diff
@@ -0,0 +1,44 @@
+package com.twitter.cassovary.algorithms.shortestpath
+
+import com.twitter.cassovary.graph.{Node, GraphDir, BreadthFirstTraverser, DirectedGraph}
+
+case class SingleSourceShortestPath(graph: DirectedGraph[Node], source: Int) extends ShortestPath {
```
pankajgupta: why case class?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  val bfs = new BreadthFirstTraverser(graph, GraphDir.OutDir, Seq(source))
+  private val (nodes, depths)  = (bfs.toList, bfs.allDepths)
+  private val previous = nodes.map { n =>
```
pankajgupta: What is the intent of this map?
Can bfsPrevNbrCounter in BFS help?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
```diff
@@ -0,0 +1,44 @@
...
+    else {
+      val newStack = currStack.zipWithIndex.map { case (seq, in) => if (in == currTop - 1) Seq(seq(0), seq(1) + 1) else seq}
+      (newStack, currTop - 1)
+    }
+    (stack, top, path)
+  }
+
+  def shortestPaths(target: Int): Paths = {
+    lazy val stream: Stream[(Stack, Int, Path)] =
+      (List(Seq(target, 0)), 0, Seq.empty[Int]) #:: stream.map { case (s, t, p) => walkOnGraph(s,t) }
```
pankajgupta: any advantage of stream over doing a 'while' loop or a 'for' comprehension ?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    stream.takeWhile{ case (_, t, _) => t >= 0}.toList
+      .filter{ case (_, _, p) => p.nonEmpty }
+      .map{ _._3 }.toSeq
```
pankajgupta: can you explain what walkOnGraph() is trying to do?

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentralitySpec.scala
```diff
@@ -0,0 +1,46 @@
...
+
+      bc(graph.getNodeById(10).get) should be (5.000  +- .005)
+      bc(graph.getNodeById(11).get) should be (2.666  +- .005)
+      bc(graph.getNodeById(12).get) should be (0.666  +- .005)
+      bc(graph.getNodeById(13).get) should be (0.666  +- .005)
+      bc(graph.getNodeById(14).get) should be (11.000 +- .005)
+      bc(graph.getNodeById(15).get) should be (11.000 +- .005)
+    }
+  }
+}
```
pankajgupta: need a test for shortest path as well

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
bmckown: The intent of this map is to keep track of immediate previous neighbors.  I think I had originally tried to use `bfs.prevNbrCounter`but, in all honesty, I found it a little difficult to get the information out of in a way that made sense.  Maybe you could offer some guidance?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
bmckown: I used a stream in an attempt to keep everything functional.  Personally, I like to avoid `var`s whenever possible so, in this case, I used a `stream` to avoid mutating the state of a given variable.  I probably could have also used a `for` comprehension.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
bmckown: No "good" reason.   I tend to get a little case class happy--I like having the option of using the object constructor and having access to all constructor arguments by default.  However, like I said, there is no "good" reason.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
bmckown: `walkonGraph()` is probably the most opaque piece of code here.  Essentially, I am trying to efficiently determine _all_ shortest paths from node i to node j--not just _a_ shortest path from node i to node j.  

Here is the output from the stack manipulation when determining the paths from node 10 to node 14 in `TestGraphs.g6`.  The result we are after is a set of paths [(10,11,14), (10,12,14), (10,13,14)].  The first line is the state of the current stack, the second line is the value of current "top"--which is essentially how far we have travelled along a given path (which could include traveling backward).
```scala
List(List(14, 0))
0
List(List(14, 0), List(11, 0))
1
List(List(14, 0), List(11, 0), List(10, 0))
2
List(List(14, 0), List(11, 1), List(10, 0))
1
List(List(14, 1), List(11, 1), List(10, 0))
0
List(List(14, 1), List(12, 0), List(10, 0))
1
List(List(14, 1), List(12, 0), List(10, 0))
2
List(List(14, 1), List(12, 1), List(10, 0))
1
List(List(14, 2), List(12, 1), List(10, 0))
0
List(List(14, 2), List(13, 0), List(10, 0))
1
List(List(14, 2), List(13, 0), List(10, 0))
2
List(List(14, 2), List(13, 1), List(10, 0))
1
List(List(14, 3), List(13, 1), List(10, 0))
0
```
So, we keep track of a list of tuples(node, index).  Index is probably a bad name for the second part as it really keeps track of the number of shortest paths that pass through that given node.  So, we keep walking starting at target and ending at source and keep visiting every node.  And at each node we visit, we ask "Is there another node that will lead me to source?"  If so, we follow that path, if not, we walk backward further until we do reach a node that could lead to another shortest path.  

Just to be illustrative, here is the output for all shortest paths from node 10 to node 15.
```scala
List(List(15, 0))
0
List(List(15, 0), List(14, 0))
1
List(List(15, 0), List(14, 0), List(11, 0))
2
List(List(15, 0), List(14, 0), List(11, 0), List(10, 0))
3
List(List(15, 0), List(14, 0), List(11, 1), List(10, 0))
2
List(List(15, 0), List(14, 1), List(11, 1), List(10, 0))
1
List(List(15, 0), List(14, 1), List(12, 0), List(10, 0))
2
List(List(15, 0), List(14, 1), List(12, 0), List(10, 0))
3
List(List(15, 0), List(14, 1), List(12, 1), List(10, 0))
2
List(List(15, 0), List(14, 2), List(12, 1), List(10, 0))
1
List(List(15, 0), List(14, 2), List(13, 0), List(10, 0))
2
List(List(15, 0), List(14, 2), List(13, 0), List(10, 0))
3
List(List(15, 0), List(14, 2), List(13, 1), List(10, 0))
2
List(List(15, 0), List(14, 3), List(13, 1), List(10, 0))
1
List(List(15, 1), List(14, 3), List(13, 1), List(10, 0))
0
```
As you can see here, we only "walk" back to 15 once since we only need to walk back to 14 before we find another node that will lead to a shortest path.  I often find this output to be the most helpful in understanding why and how this algorithm works.  As per usual, I followed the example from networkx for shortest paths.  I will try to find the source.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+trait ShortestPath {
+  type Path  = Seq[Int]
+  type Paths = Seq[Path]
+  type Stack = List[Path]
+
+  def graph: DirectedGraph[Node]
+
+  /**
+   * Calculate the shortest path to a given target
```
pankajgupta: From which node? All nodes? Might want to clarify

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,40 @@
...
+   * @param target The integer ID of the target node
+   * @return A sequence of paths.  If no shortest path exists, the list will be empty
+   */
+  def shortestPaths(target: Int): Paths
```
pankajgupta: Minor: I will suggest not aliasing Paths at all and leaving it as Seq[Path]. This makes it more readable.

pankajgupta: Also, suggest renaming this to 'shortestPathsTo' to make it a bit clearer that we are supplying a target (and not a source).

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
pankajgupta: Also s/path/paths because there might be multiple shortest paths

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
pankajgupta: Talking just about walkOnGraph(), I don’t understand it fully but I think what you are doing is retracing back from target node and keeping track of ancestor in ‘index’. My first question is why you are tracing backwards (which relies on ‘previous’) and not moving forwards out of source node along its outneighbors, in which case this can be much simpler as a variation of BFS where you keep track of parent of node as well as the shortest path length. You enqueue the node only if the current depth is equal or less than the shortest path length value currently stored in the node.
Taking g6 as an example, you would start from source 10 say, enqueue 11,12,13 and keep 10 as ancestor for each of them as well as record a value of ‘1’ for the shortest path length to them. Then consider node 11 which has two children 12 and 14:

You will enqueue 14 (together with ancestor=11, path length=2) because no shortest path length has been stored with 14 so far.

You will not enqueue 12 but already a shorter path length of 1 has been stored.

This can then be coded just like BFS as an extension of queuebased traverser in Traverser.scala which would be the preferred way of doing the all shortest paths calculation.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
pankajgupta: Stream is useful if you want lazy evaluation of the tail, which you are not doing here because you are evaluating takeWhile immediately. I think you will be better off with a while or a for loop here. vars are good to avoid, but making code much harder to read is not a good tradeoff i think.

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPathSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+ */
+package com.twitter.cassovary.algorithms.shortestpath
+
+import com.twitter.cassovary.graph._
+import org.scalatest.{Matchers, WordSpec}
+
+class SingleSourceShortestPathSpec extends WordSpec with Matchers {
+  lazy val graph = TestGraphs.g6
+
+  "Single source shortest path" should {
```
pankajgupta: should probably test on a couple more simple graphs and few more source nodes

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/centrality/BetweennessCentrality.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+package com.twitter.cassovary.algorithms.centrality
+
+import com.twitter.cassovary.algorithms.shortestpath.SingleSourceShortestPath
+import com.twitter.cassovary.graph._
+
+class BetweennessCentrality(graph: DirectedGraph[Node], normalize: Boolean = true) extends AbstractCentrality(graph) {
+
+  private def combineMap(a: Map[Int, Double], b: Map[Int, Double]): Map[Int, Double] = {
+    (a.keys ++ b.keys).map { k => k -> (a.getOrElse(k, 0.0) + b.getOrElse(k, 0.0)) }.toMap
```
pankajgupta: s/.keys/.keySet to be explicit

bmckown:
Updated.  @pankajgupta please take a look

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import com.twitter.cassovary.graph._
+
+/**
+ * Starting from a given source, calculate the shortest paths from that source to
+ * a given target node.  This algorithm returns all shortest paths available not just
+ * a single shortest path.
+ * @param source The source node with which all shortest paths start.
+ */
+class SingleSourceShortestPath(val graph: DirectedGraph[Node], val source: Int) extends ShortestPath {
```
pankajgupta: I'd merge this into the trait ShortestPath and just have this class SingleSourceShortestPath rather than have an unnecessary trait lying around.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,44 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.algorithms.shortestpath
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
+
+/**
+ * A base trait for all ShortestPath algorithm implementations
+ */
+trait ShortestPath {
```
pankajgupta: it really is now single source shortest path, so I'd merge this into the class (also see below).

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/ShortestPath.scala
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Source node for all shortest paths calculated by `shortestPaths`
+   */
+  val source: Int
+
+  /**
+   * Calculate the shortest paths to `target` from a given source node.  The source
+   * node must be provided as an instance variable of all implementing subclasses.
+   * @param target The integer ID of the target node
+   * @return A sequence of paths.  If no shortest path exists, the list will be empty
```
pankajgupta: Are these paths ordered by some thing? If not, may be return a Set? If yes, specify in the comment how they are ordered.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -248,7 +252,7 @@ trait QueueBasedTraverser[+V <: Node] extends Traverser[V] {
   def nodePriority: GraphTraverserNodePriority
 
   /**
-   * Action performed when visiting node `node` (before iteratur retuns `node`).
+   * Action performed when visiting node `node` (before iterator returns `node`).
```
pankajgupta: thanks for fixing the spelling!

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
     super.visitNode(node)
   }
 
-  def shouldBeEnqueued(color: Walk.NodeColor.Color): Boolean = {
-    color == Walk.NodeColor.Unenqueued
+  def shouldBeEnqueued(color: Color): Boolean = color == Walk.NodeColor.Unenqueued
```
pankajgupta: I'd keep the {} here.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+}
+
+/**
+ * Traverses in BFS order and keeps track of all shortest paths from any node in `homeNodeIds` to
+ * all target paths.  The traversal happens along all outbound nodes, it is suggested that `graph`
+ * storage direction is `GraphDir.OutDir`
+ * @param graph The graph to traverse on.  This graph need not necessarily be directed, but the
+ *              traversal will happen along all outbound nodes.
+ * @param homeNodeIds The sequence of source nodes to find all shortest paths.  Each shortest path must
```
pankajgupta: s/find all shortest paths/find all shortest paths from/

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+                                        val dir: GraphDir = GraphDir.OutDir,
+                                        walkLimits: Walk.Limits = Walk.Limits())
+  extends DepthTracker with BoundedIterator[V] {
+
+  // Queueing parameters
+  override def nodePriority = GraphTraverserNodePriority.FIFO
+  override lazy val limits = walkLimits
+
+  // Accounting parameters
+  val numPaths = mutable.Map.empty[Int, Int]
```
pankajgupta: use IntInfoKeeper instead? It is meant to be more efficient as it uses fastutils.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -324,6 +328,10 @@ trait QueueBasedTraverser[+V <: Node] extends Traverser[V] {
     }
   }
 
+  protected def chooseNodes(node: Node): Seq[Int] = {
```
pankajgupta: perhaps have a more descriptive name? may be chooseNeighborsAfterVisit() or something similar...

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+    numPaths(node.id) = if (numPaths.contains(node.id)) numPaths(node.id) + 1 else 1
+
+    val currDepth = depth(node.id).get
+    val nextDepth = currDepth + 1
+
+    val index = numPaths(node.id) - 1
+
+    node.neighborIds(dir) filter { out =>
+      val shortestPathLength = depth(out)
+      val update = paths(node.id)(index) ++ mutable.Seq(node.id)
```
pankajgupta: why mutable?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+
+      if (shortestPathLength.isDefined) {
+        if (nextDepth <= shortestPathLength.get) {
+          paths(out) = paths(out) :+ update
+          true
+        } else false
+      }
+      else {
+        paths(out) = mutable.Seq(update)
```
pankajgupta: why mutable?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+   * In the filtering process, the shortest paths will be updated simultaneously.
+   * @return The next set of nodes to be enqueued
+   */
+  override protected def chooseNodes(node: Node): Seq[Int] = {
+    numPaths(node.id) = if (numPaths.contains(node.id)) numPaths(node.id) + 1 else 1
+
+    val currDepth = depth(node.id).get
+    val nextDepth = currDepth + 1
+
+    val index = numPaths(node.id) - 1
```
pankajgupta: looks like you are only accessing the tail of paths(node.id) using 'index', and if so no need to keep numPaths(node.id) at all, yes?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+
+    node.neighborIds(dir) filter { out =>
+      val shortestPathLength = depth(out)
+      val update = paths(node.id)(index) ++ mutable.Seq(node.id)
+
+      if (shortestPathLength.isDefined) {
+        if (nextDepth <= shortestPathLength.get) {
```
pankajgupta: don't think it can ever be '<'. Equal to is possible but not less than, yes? If so, might want to assert on it.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -442,9 +448,67 @@ class BreadthFirstTraverser[+V <: Node](val graph: Graph[V], val dir: GraphDir,
...
+      else {
+        paths(out) = mutable.Seq(update)
+        true
+      }
+    }
   }
+
+  //Always enqueue node independent of color
+  override def shouldBeEnqueued(color: Color): Boolean = true
 }
```
pankajgupta: what you have works, but is exponential in both time and space. What do you think about just keeping the information of 'seq of ancestors of a node'. So, in TestGraph.g6 if source node is 10. ancestors(10) = Nil, ancestors(11) = Seq(10), ancestors(12) = Seq(10), ancestors(13) = Seq(10), ancestors(14) = Seq(11, 12, 13), ancestors(15) = Seq(14). By the nature of BFS, I think it will work to not enqueue a neighbor n if it has been enqueued before, but append node.id to ancestors(n).

The full set of paths to a target node t will then be recursively the set of paths to ancestors(t) with 't' appended to each of them. This can be exponential but can be computed if/when needed by some other code, but at least not at the traverser level.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/shortestpath/SingleSourceShortestPath.scala
```diff
@@ -0,0 +1,31 @@
...
+  private val spt = new ShortestPathsTraverser(graph, Seq(source))
+
+  //We need to run the iterator...
+  spt.toList
```
pankajgupta: perhaps put in a lazy val to not have to run unless needed?

pankajgupta:
How is this going @bmckown ?

bmckown:
@pankajgupta I'm currently working on it.  The calculation is actually remarkably slow for large graphs.  I'm doing a little bit of work on parallelizing the calculation in hopes of speeding it up as much as possible.  Even the implementations I've played with in other libraries are pretty slow.

bmckown:
I'm going to submit a new PR to clean some of this up.

PR #167, opened 2015-02-25
Add linkanalysis package and implement Hits

bmckown:
An initial attempt at incorporating an additional link analysis algorithm (HITS).  I have created a new `linkanalysis` package with a base `LinkAnalysis` trait as well as an `AbstractLinkAnalysis` base class.  This helped to abstract much of the overlap between PageRank and HITS.  Also, I am currently doing the iteration by using a Stream and finding the first iteration that violates an iteration or threshold condition using find.  If anyone has a better idea of how to do this (I like to avoid `var`s when possible), I am all ears.  Thanks!


cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+   */
+  val error = result.error
+
+  /**
+   * The final iteration number our algorithm terminated at
+   */
+  val iterations = result.iteration
+
+  def run(init: T = start): T = {
+    lazy val stream: Stream[T] = init #:: stream map { h => iterate(h)}
```
pankajgupta: Because of memoization, this will cause stream to hold on to all values in the stream. Not sure if you want to, or discard earlier values. Streams are generally tricky: see http://blog.dmitryleskov.com/programming/scala/stream-hygiene-i-avoiding-memory-leaks/

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+    val progress = Progress(s"${modelName}_init", 65536, Some(graph.nodeCount))
+
+    stream.find { s =>
+      log.debug("Beginning %sth iteration".format(s.iteration))
+      progress.inc
+      if (maxIterations.isDefined)
+        s.iteration >= maxIterations.get || s.error <= tolerance
+      else
+        s.error <= tolerance
+    }.get
```
pankajgupta: streams are overkill for this. A simple recursive method will do here. i think what you are trying to do is:

def loop(curr: T, i: Int): T = { 
val next = iterate(curr)
if (i < maxIterations && next.error > tolerance) loop(next, i+1) else next
}

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+import com.twitter.cassovary.util.Progress
+import com.twitter.logging.Logger
+
+/**
+ * Base trait for link analysis algorithms.  The goal of these is to iteratively determine the importance
+ * of the nodes in the graph by analyzing how each of those nodes is connected to its neighbors.
+ * @tparam T `LinkAnalysis` must be generically typed by `Iteration` or one of its subclasses.  An `Iteration`
+ *          holds all of the pertinent information for a given algorithm.
+ */
+trait LinkAnalysis[T <: Iteration] extends (Option[T] => T) {
```
pankajgupta: The 'extends (Option[T] => T)' looks artificial. I'd remove it and let the user use run(). Also I'd merge the trait LinkAnalysis and abstract class AbstractLinkAnalaysis to one abstract class.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+ * @param tolerance The maximum error allowed.
+ */
+abstract class Params(val maxIterations: Option[Int], val tolerance: Double)
+
+/**
+ * The base class for all iterations through our iterative algorithms.  These classes will simply hold all of the
+ * information needed to assess the number of iterations, the error, and the current set of values.
+ * @param error  The T1 error for the current iteration vs the previous iteration.
+ */
+abstract class Iteration(val error: Double, val iteration: Int)
```
pankajgupta: consider calling "iteration" "number" or "iterationNumber" else the element 'iteration' can be confused with the class

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+    if (t1) rawError else Math.sqrt(rawError)
+  }
+}
+
+/**
+ * The base class for all parameters fed to our iterative algorithms.
+ * @param maxIterations The maximum number of times that the link analysis algorithm will run before termination
+ * @param tolerance The maximum error allowed.
+ */
+abstract class Params(val maxIterations: Option[Int], val tolerance: Double)
```
pankajgupta: change these vals to defs to not have subclasses to 'override' them all the time

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
```diff
@@ -0,0 +1,70 @@
...
+                             override val iteration: Int)
+  extends Iteration(error, iteration)
+
+/**
+ * PageRank is a link analysis algorithm designed to measure the importance of nodes in a graph.
+ * Popularized by Google.
+ *
+ * Unoptimized for now, and runs in a single thread.
+ */
+case class PageRank(graph: DirectedGraph[Node], params: PageRankParams)
```
pankajgupta: case classes are great for small classes such as Iteration or Params, but I don't generally use case classes unless there is a clear reason to for classes with more code / elements.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
pankajgupta: Or remove the 'val' annotation

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+ * @tparam T `LinkAnalysis` must be generically typed by `Iteration` or one of its subclasses.  An `Iteration`
+ *          holds all of the pertinent information for a given algorithm.
+ */
+trait LinkAnalysis[T <: Iteration] extends (Option[T] => T) {
+
+  /**
+   * Run the algorithm using an optionally provided `init` argument or with the the default argument provided by start
+   * @param init An initial iteration
+   */
+  def apply(init: Option[T] = None): T = if (init.isDefined) run(init.get) else run()
```
pankajgupta: remove. Adds to the confusion between initial/start/init.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
pankajgupta: Or s/Iteration/IterationState

pankajgupta: Does this need to be exposed outside AbstractLinkAnalysis, or can it be completely internal?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
```diff
@@ -0,0 +1,70 @@
...
+    PageRankIteration(initial, 100 + tolerance, 0)
+  }
+
+  def iterate(prevIteration: PageRankIteration): PageRankIteration = {
+    val beforePR = prevIteration.pageRank
+    val afterPR  = new Array[Double](graph.maxNodeId + 1)
+
+    log.debug("Calculating new PageRank values based on previous iteration...")
+    val prog = Progress("pagerank_calc", 65536, Some(graph.nodeCount))
+    graph.foreach { node =>
```
pankajgupta: don't need the '.' after graph

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+  protected val log = Logger.get(modelName)
+
+  val maxIterations = params.maxIterations
+  val tolerance     = params.tolerance
+
+  /**
+   * This is the final `Iteration` object from our stream.  The user can feed this iteration back to `iterate`
+   * or `run` if they wish to converge further.
+   */
+  val result = if ((maxIterations.isDefined && maxIterations.get > 0) || !maxIterations.isDefined)
```
pankajgupta: not sure what the benefit of evaluating result, error, iteration is. let the user call run() and get the final error and iterations used?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,121 @@
...
+   * @return An default starting iteration.
+   */
+  protected def start: T
+
+  /**
+   * Calculate the error between two arrays using either the T1 error or the T2 error
+   * @param t1 Flag `true` to calculate the T1 error (the sum of absolute differences between two arrays).  Flag
+   *           `false` to calculate the T2 error (the sum of the squared differences between two arrays).
+   */
+  protected def error(a: Array[Double], b: Array[Double], t1: Boolean = true): Double = {
```
pankajgupta: probably call this delta than error? Also, a bit weird to have T everywhere but make an Array[] specific method here. Guess it is a convenience method in which case add so to the comment and perhaps call it 'deltaOfArrays' and not a generic 'error'

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRankSpec.scala
```diff
@@ -32,33 +32,45 @@ class PageRankSpec extends WordSpec with Matchers {
...
+  lazy val graphG6 = TestGraphs.g6
 
-    lazy val graphG6 = TestGraphs.g6
+  "PageRank" should {
 
     "Return a uniform array with 0 iterations" in {
       val params = PageRankParams(0.1, Some(0))
       val pr = PageRank(graphG6, params)
-      pr should almostEqualMap(Map(10 -> 1.0/6, 11 -> 1.0/6, 12 -> 1.0/6, 13 -> 1.0/6, 14 -> 1.0/6, 15 -> 1.0/6))
+      pr.iterations shouldEqual 0
```
pankajgupta: why 0?

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PageRankBenchmark.scala
```diff
@@ -24,4 +24,4 @@ class PageRankBenchmark(graph: DirectedGraph[Node],
   def operation() {
     PageRank(graph, pageRankParams)
   }
-}
\ No newline at end of file
+}
```
pankajgupta: can you report on the pagerank benchmark results before and after your change on a couple of graphs?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,75 @@
...
+    
+    graph foreach { n => hubs(n.id) = 1.0 / graph.nodeCount }
+    new HitsIteration(hubs, authorities, 0, 100 + tolerance)
+  }
+  
+  def iterate(prevIteration: HitsIteration): HitsIteration = {
+    val (beforeHubs, beforeAuth) = (prevIteration.hubs, prevIteration.authorities)
+
+    val rawAuth = beforeAuth.zipWithIndex.foldLeft(new Array[Double](graph.maxNodeId + 1)) { case (partialAuth, (_, node)) =>
+      if (graph.existsNodeId(node)) {
```
pankajgupta: perhaps filter out all nodes that exist outside the iterate loop?
val idsInGraph = graph filter existsNodeId

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,75 @@
...
+        graph.getNodeById(node).get.outboundNodes().foreach { nbr => partialAuth(nbr) += beforeHubs(node)}
+      }
+      partialAuth
+    }
```
pankajgupta: the zipWithIndex and foldLeft is unnecessary as you are just iterating over all node ids.
val partialAuth = new Array[Double](maxNodeId+1)
for (id <- graph) { graph.getNodeById(id).get.outboundNodes() foreach {...} }

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,75 @@
...
+    val rawHubs = beforeHubs.zipWithIndex.map { case (_, node) =>
+      if(graph.existsNodeId(node))
+        graph.getNodeById(node).get.outboundNodes().foldLeft(0.0) { (partialSum, nbr) => partialSum + rawAuth(nbr) }
+      else 0.0
+    }
```
pankajgupta: need to be a bit careful with the direction stored in the graph. outboundNodes() is cheap if graph.IsDirStored(OutDir) is true, else we need to use inboundNodes(). Furthur, we can use inboundNodes() when isDirStored(InDir) is true. This is also true for the pagerank calculation that was ignored in the existing code as well.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,75 @@
...
+ * Case class for performing Hits algorithm.  Hits is a link analysis algoritm that returns two values characterizing
+ * each node.  Each node receives both a hub value and an authority value.  A node that is characterized by a large
+ * hub value is one that has many high quality outbound connections to other nodes, while a node that is characterized
+ * by a large authority value has inbound connections from high quality hub nodes.
+ * @param params The set of all parameters passed into our algorithm
+ */
+case class Hits(graph: DirectedGraph[Node], params: HitsParams)
+  extends AbstractLinkAnalysis[HitsIteration](graph, params, "hits") {
+  
+  val normalize = params.normalize
```
pankajgupta: want to make this public?

pankajgupta:
btw, also look at IterativeLinkAnalyzer in com.twitter.cassovary.graph.bipartite which allows running an iterative algorithm similar to HITS on a bipartite graph.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PageRankBenchmark.scala
bmckown: After trying to figure out how to run benchmarks for a while, I decided to just ask.  How do I run the benchmarks?  Thanks!

I will probably update the README with instructions on how to run them, unless of course I'm just being an idiot.

pankajgupta: Just added cassovary-benchmarks/BenchmarksREADME.md -- please feel free to
modify and look at the individual sources of benchmarks as well.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
bmckown: You're correct.  Using a `stream` caused a large decrease in performance (page rank was taking 3x longer to converge).

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PageRankBenchmark.scala
bmckown: Here are some benchmarks.  I ran each 1000 times.  Here are the benchmarks for the code that is currently in master:

<table>
  <tr>
    <td></tb><td><b>Personalized PR</b></td><td><b>Global PR</b></td>
  <tr>
    <td><b>Facebook Graph</b></tb><td>
0 ms 969 &mu;s</td><td>8 ms 52 &mu;s</td>
  </tr>
  <tr>
    <td><b>Wiki Graph</b></tb><td>1 ms 134 &mu;s</td><td>14 ms 778 &mu;s</td>
  </tr>
</table>

And affter adding the `linkanalysis` package.

<table>
  <tr>
    <td></tb><td><b>Personalized PR</b></td><td><b>Global PR</b></td><td><b>HITS</b></td>
  <tr>
    <td><b>Facebook Graph</b></tb><td>
1 ms 11 &mu;s</td><td>10 ms 542 &mu;s</td><td>107 ms 601 &mu;s</td>
  </tr>
  <tr>
    <td><b>Wiki Graph</b></tb><td>1 ms 247 &mu;s</td><td>18 ms 517 &mu;s</td><td>81 ms 611 &mu;s</td>
  </tr>
</table>

bmckown:
To keep this comment from getting hidden by updates:

pankajgupta:
Any ideas why there is a small regression? Almost 25% in the case of
GlobalPR?

bmckown:
I did some investigating.  I think the regression comes from calculating the `error` in each iteration found [here](https://github.com/twitter/cassovary/pull/167/files#diff-4582d490c3275618c123d9d6c4e01743R63).  So I removed the error calculation and reran the benchmarks.  Here are the results:

Global PR Performance Results
-------------------------------------------
<table>
  <tr>
    <td></tb><td><b>Master</b></td><td><b>With Error Calculation</b></td><td><b>Without Error Calculation</b></td>
  <tr>
    <td><b>Facebook Graph</b></tb><td>8 ms 52 &mu;s</td><td>9 ms 696 &mu;s</td><td>7 ms 868 &mu;s</td>
  </tr>
  <tr>
    <td><b>Wiki Graph</b></tb><td>14 ms 778 &mu;s</td><td>18 ms 492 &mu;s</td><td>15 ms 542 &mu;s</td>
  </tr>
</table>

pankajgupta:
That time looks better. Will look at your updated review next.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+import com.twitter.cassovary.graph.{StoredGraphDir, Node, DirectedGraph}
+import com.twitter.cassovary.util.Progress
+import com.twitter.logging.Logger
+
+import scala.annotation.tailrec
+
+/**
+ * The base class for all parameters fed to our iterative algorithms.
+ */
+abstract class Params { def maxIterations: Option[Int]; def tolerance: Double }
```
pankajgupta: More natural to have a non-trivial class split into multiple lines

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * The base class for all iterations through our iterative algorithms.  These classes will simply hold all of the
+ * information needed to assess the number of iterations, the error, and the current set of values.
+ */
+abstract class IterationState { def error: Double; def iteration: Int }
```
pankajgupta: Ditto about multiple lines

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * `AbstractLinkAnalysis` is the base class that extends `LinkAnalysis`.  All algorithms should inherit from
```
pankajgupta: Update comment

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+    params: Params, modelName: String) {
+
+  protected val log = Logger.get(modelName)
+
+  protected val maxIterations = params.maxIterations
+  protected val tolerance     = params.tolerance
+
+  /**
+   * Run a single iteration through our algorithm.
+   * @param iteration The starting iteration that our algorithm will be applied to.
```
pankajgupta: s/iteration/currState

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+   * @return A new iteration.
+   */
+  def iterate(iteration: T): T
+
+  /**
+   * Provides default initial start values for our algorithms.
+   * @return An default starting iteration.
+   */
+  protected def defaultInitialIteration: T
```
pankajgupta: s/defaultInitialIteration/defaultInitialState/

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+      log.info("Warning - you may be able to reduce the memory usage by renumbering this graph!")
+
+    log.debug(s"Initializing starting ${modelName}...")
+    val progress = Progress(s"${modelName}_init", 65536, Some(graph.nodeCount))
+
+    def terminationCondition: Boolean = if (maxIterations.isDefined)
+      iters < maxIterations.get && error > tolerance
+    else error > tolerance
+
+    while (terminationCondition) {
```
pankajgupta: minor code change to minimize the number of vars would be to 
1. remove the vars 'iters' and 'error'.
2. s/def terminationCondition/def terminate(currState)/
3. s/while (terminationCondition)/while (!terminate(currentIteration)/

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,74 @@
+package com.twitter.cassovary.algorithms.linkanalysis
+
+import com.twitter.cassovary.graph.{Node, DirectedGraph}
```
pankajgupta: alphabetical order

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/LinkAnalysis.scala
```diff
@@ -0,0 +1,97 @@
...
+ * this class.  Let it be noted that we exploit scala streams to accomplish the recursion necessary for
+ * our algorithms.  We will not be storing the results every step of the way--instead we find the iteration
+ * that first satisfies the condition of exceeding maximum iterations or falls below the error threshold.
+ * @param params The set of all parameters passed into our algorithm
+ * @param modelName This must be provided by all subclasses for logging purposes.  This should not be
+ *                  left up to the user to define.  This must be provided by the implementer.
+ * @tparam T `LinkAnalysis` must be generically typed by `Iteration` or one of its subclasses.  An `Iteration`
+ *           holds all of the pertinent information for a given algorithm.
+ */
+abstract class AbstractLinkAnalysis[T <: IterationState](graph: DirectedGraph[Node],
```
pankajgupta: consider s/AbstractLinkAnalysis/LinkAnalysis/

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,74 @@
...
+  private def scale(m: Array[Double], byMax: Boolean): Array[Double] = {
+    val d = if (byMax) m.max else m.sum
+    if (d <= 0) m else m.map { v => v / d }
+  }
+
+  protected def defaultInitialIteration: HitsIterationState = {
+    val hubs = new Array[Double](graph.maxNodeId + 1)
+    val authorities = new Array[Double](graph.maxNodeId + 1)
+    
+    graph foreach { n => hubs(n.id) = 1.0 / graph.nodeCount }
```
pankajgupta: look at Array.fill

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,74 @@
...
+    new HitsIterationState(hubs, authorities, 0, 100 + tolerance)
+  }
+  
+  def iterate(prevIteration: HitsIterationState): HitsIterationState = {
+    val beforeHubs = prevIteration.hubs
+
+    val partialAuth, partialHubs = new Array[Double](graph.maxNodeId + 1)
+    graph foreach { node =>
```
pankajgupta: I know it hasn't been done for old code but since this is new code, we should consider the various directions the graph might be stored in: Out only / In only / Both, and (1) use it to our advantage if we can, and (2) Make sure code works efficiently in the 3 cases

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
```diff
@@ -0,0 +1,74 @@
...
+    }
+    graph foreach { node =>
+      partialHubs(node.id) = node.outboundNodes().foldLeft(0.0) { (partialSum, nbr) => partialSum + partialAuth(nbr) }
+    }
+
+    val (afterHubs, afterAuth) = (scale(partialHubs, byMax = true), scale(partialAuth, byMax = true))
+    new HitsIterationState(afterHubs, afterAuth, prevIteration.iteration + 1, deltaOfArrays(beforeHubs, afterHubs))
+  }
+
+  override def run(init: HitsIterationState = defaultInitialIteration): HitsIterationState = {
```
pankajgupta: have a def postRun(T): T = {} method instead in LinkAnalysis that is called at the end of run()

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/linkanalysis/HitsSpec.scala
```diff
@@ -0,0 +1,90 @@
+package com.twitter.cassovary.algorithms.linkanalysis
+
+import com.twitter.cassovary.graph.TestGraphs
+import org.scalatest.{Matchers, WordSpec}
+
+class HitsSpec extends WordSpec with Matchers {
+  lazy val graph = TestGraphs.g6
```
pankajgupta: Test with a graph kept only in inbound direction as well

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/Hits.scala
bmckown: @pankajgupta In working on this, I noticed something that didn't quite make sense to me--before submitting an issue, I wanted to make sure this was an actual bug.  

In TestGraphs, `g6` is created using 
```scala  
val nodeSeqIterator = Seq(
      NodeIdEdgesMaxId(10, Array(11, 12, 13)),
      NodeIdEdgesMaxId(11, Array(12, 14)),
      NodeIdEdgesMaxId(12, Array(14)),
      NodeIdEdgesMaxId(13, Array(12, 14)),
      NodeIdEdgesMaxId(14, Array(15)),
      NodeIdEdgesMaxId(15, Array(10, 11))
      )
```
which, according to the documentation, should be an iterator of `NodeIdEdgesMaxId` objects with `nodeId` as the first constructor arg and an array of out edges as the second arg (seen [here](https://github.com/twitter/cassovary/blob/master/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ArrayBasedDirectedGraph.scala#L36-L39).

So when `g6_onlyin` and `g6_onlyout` are created [here](https://github.com/twitter/cassovary/blob/master/cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala#L99-L102), they _should_ represent the same graph but only differ in node storage strategy.

In the console, if you print out `g6`, you get the following:
```scala
both: com.twitter.cassovary.graph.ArrayBasedDirectedGraph =
Node count: 6
Edge count: 11
Nodes:
NodeId => 10
InboundNodes[1] =>15|
OutboundNodes[3] =>11|12|13|

NodeId => 11
InboundNodes[2] =>10|15|
OutboundNodes[2] =>12|14|

NodeId => 12
InboundNodes[3] =>10|11|13|
OutboundNodes[1] =>14|

NodeId => 13
InboundNodes[1] =>10|
OutboundNodes[2] =>12|14|

NodeId => 14
InboundNodes[3] =>11|12|13|
OutboundNodes[1] =>15|

NodeId => 15
InboundNodes[1] =>14|
OutboundNodes[2] =>10|11|
```
`g6_onlyout` prints:
```scala
out: com.twitter.cassovary.graph.ArrayBasedDirectedGraph =
Node count: 6
Edge count: 11
Nodes:
NodeId => 10
InboundNodes[0] =>
OutboundNodes[3] =>11|12|13|

NodeId => 11
InboundNodes[0] =>
OutboundNodes[2] =>12|14|

NodeId => 12
InboundNodes[0] =>
OutboundNodes[1] =>14|

NodeId => 13
InboundNodes[0] =>
OutboundNodes[2] =>12|14|

NodeId => 14
InboundNodes[0] =>
OutboundNodes[1] =>15|

NodeId => 15
InboundNodes[0] =>
OutboundNodes[2] =>10|11|
```
which is expected.  *BUT* `g6_onlyin` has a very strange behavior:
```scala
in: com.twitter.cassovary.graph.ArrayBasedDirectedGraph =
Node count: 6
Edge count: 11
Nodes:
NodeId => 10
InboundNodes[3] =>11|12|13|
OutboundNodes[0] =>

NodeId => 11
InboundNodes[2] =>12|14|
OutboundNodes[0] =>

NodeId => 12
InboundNodes[1] =>14|
OutboundNodes[0] =>

NodeId => 13
InboundNodes[2] =>12|14|
OutboundNodes[0] =>

NodeId => 14
InboundNodes[1] =>15|
OutboundNodes[0] =>

NodeId => 15
InboundNodes[2] =>10|11|
OutboundNodes[0] =>
```
It seems as though `g6_onlyin` has just taken the outbound edges and stored them as inbound edges which means that `g6_onlyin` and `g6_onlyout` do not represent the same graph.  

Just wanted a sanity check before I submit an issue.

pankajgupta: That’s right — the iterator is seen as a sequence of inedges. It’s not a bug, perhaps the name of the variable g6_onlyin is confusing.

bmckown:
All results are after 250 reps.

Graphs stored in outbound direction
------------------------------------------------
<table>
  <tr>
    <td></tb><td><b>Personalized PR</b></td><td><b>Global PR</b></td><td><b>HITS</b></td>
  <tr>
    <td><b>Facebook Graph</b></tb><td>
1 ms 112 &mu;s</td><td>7 ms 733 &mu;s</td><td>93 ms 482 &mu;s</td>
  </tr>
  <tr>
    <td><b>Wiki Graph</b></tb><td>1 ms 135 &mu;s</td><td>11 ms 399 &mu;s</td><td>64 ms 147 &mu;s</td>
  </tr>
</table>

Graphs stored in inbound direction
----------------------------------------------
<table>
  <tr>
    <td></tb><td><b>Personalized PR</b></td><td><b>Global PR</b></td><td><b>HITS</b></td>
  <tr>
    <td><b>Facebook Graph</b></tb><td>
1 ms 129 &mu;s</td><td>8 ms 376 &mu;s</td><td>106 ms 597 &mu;s</td>
  </tr>
  <tr>
    <td><b>Wiki Graph</b></tb><td>1 ms 251 &mu;s</td><td>12 ms 224 &mu;s</td><td>79 ms 39 &mu;s</td>
  </tr>
</table>

bmckown:
I'll solve this merge conflict and then submit a fresh PR

PR #173, opened 2015-03-10
Add centrality performance benchmarks

bmckown:
After 500 iterations for in-degree, and out-degree, and 10 iterations for closeness:
<table>
  <tr>
    <td></td><td><b>In-Degree</b></td><td><b>Out-Degree</b></td><td><b>Closeness</b></td>
  </tr>
  <tr>
    <td><b>Facebook</b></td><td>268 &mu;s</td><td>348 &mu;s</td><td>3 s 521 ms</td>
  </tr>
  <tr>
    <td><b>Wiki</b></td><td>238 &mu;s</td><td>241 &mu;s</td><td>17 s 514 ms</td>
  </tr>
</table>
The degree centrality calculations are very fast, however, the closeness centrality calculation is rather slow.  We should try to find some potential optimizations in the future.

Once betweenness centrality has been merged, I will add a benchmark for that as well.

### DIFF
--- a/cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
+++ b/cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
@@ -85,7 +85,7 @@ object PerformanceBenchmark extends App with GzipGraphDownloader {
   val globalPRFlag = flags("globalpr", false, "run global pagerank benchmark")
   val pprFlag = flags("ppr", false, "run personalized pagerank benchmark")
   val centFlag = flags("c", DEFAULT_CENTRALITY_ALGORITHM,
-    "run the specified centrality algorithm (indegree, outdegree, closeness)")
+    "run the specified centrality algorithm (indegree, outdegree, closeness, all)")
   val getNodeFlag = flags("gn", 0, "run getNodeById benchmark with a given number of steps")
   val reps = flags("reps", DEFAULT_REPS, "number of times to run benchmark")
   val adjacencyList = flags("a", false, "graph in adjacency list format")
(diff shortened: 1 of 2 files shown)


cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
```diff
@@ -83,6 +83,9 @@ object PerformanceBenchmark extends App with GzipGraphDownloader {
   val helpFlag = flags("h", false, "Print usage")
   val globalPRFlag = flags("globalpr", false, "run global pagerank benchmark")
   val pprFlag = flags("ppr", false, "run personalized pagerank benchmark")
+  val idcFlag = flags("idc", false, "run in-degree centrality benchmark")
```
pankajgupta: In the interest of not bloating the command args, I'll suggest doing this hierarchically, i.e., "-c" could denote that we should run a centrality benchmark and the followon value could be one of "indegree" or "outdegree" or "closeness" or "all".

pankajgupta:
Filing a new issue to improve the performance of closeness centrality

bmckown:
Updated.  Let me know what you think.  Thanks for filing the issue--that calculation is really rather slow.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
```diff
@@ -83,6 +84,8 @@ object PerformanceBenchmark extends App with GzipGraphDownloader {
   val helpFlag = flags("h", false, "Print usage")
   val globalPRFlag = flags("globalpr", false, "run global pagerank benchmark")
   val pprFlag = flags("ppr", false, "run personalized pagerank benchmark")
+  val centFlag = flags("c", DEFAULT_CENTRALITY_ALGORITHM,
+    "run the specified centrality algorithm (indegree, outdegree, closeness)")
```
pankajgupta: add all

pankajgupta:
BTW, we use yourkit.com to do profiling if interested.

PR #175, opened 2015-03-12
Added a splitter to split input files of a graph into multiple subdirectories given a partitioner

pankajgupta:


pankajgupta:
CC @szymonm

PR #177, opened 2015-03-12
Remove obsolete threadpool and executor service.

pankajgupta:
CC @szymonm 

After the export to future pool, the old executor service is no longer needed.


pankajgupta:
Ping @szymonm

pankajgupta:
Needed this so merging for now. Please still take a look when you can @szymonm

szymonm:
You are right. That's not needed anymore.

PR #181, opened 2015-04-21
Ask Travis CI to retry resolution and downloads

travisbrown:
@pankajgupta We're investigating the `maven.twttr.com` problems internally, but I wanted to see whether this workaround (which we're using in other projects) will fix the issue here. It only retries the resolution and downloading step, so it won't slow down notifications on pull requests with "real" failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,8 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 
+before_script:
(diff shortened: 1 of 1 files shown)


pankajgupta:
Thanks @travisbrown for the patch!

pankajgupta:
(The patch itself did not pass CI with all the jvm options, so looking forward to seeing whether the maven.twttr.com problems can themselves be solved.)

Issue #186, opened 2015-06-17
[Bug?] PageRank sum is much less 1


alexloginov:
Hi Everybody!

I'm trying to calculate PageRank for a graph [toy_7nodes_adj_StringIds.txt](https://github.com/twitter/cassovary/blob/master/cassovary-core/src/test/resources/graphs/toy_7nodes_adj_StringIds.txt) using cassovary-core_2.11:5.0.0

My code is pretty simple:
```scala
package com.datanovo.pagerank

import com.twitter.cassovary.algorithms.{PageRankParams, PageRank}
import com.twitter.cassovary.util.SequentialNodeNumberer
import com.twitter.cassovary.util.io.{AdjacencyListGraphReader}

object FailingPR {
  def main(args: Array[String]): Unit = {
    val numberer = new SequentialNodeNumberer[String]()
    val graph = new AdjacencyListGraphReader[String](
      ".",
      "toy_7nodes_adj_StringIds.txt",
      numberer,
      idReader = identity
    ).toArrayBasedDirectedGraph()

    val result = PageRank(graph, PageRankParams(dampingFactor = 0.85))
    println (result.sum)
  }
}

```

But I was very surprised with results:
> 0.2810312946428572

So it is not 1 in total. After this i decided to print out ranks for each node separately:

>a:0.03266071428571429
b:0.06042232142857144
c:0.047108058035714294
d:0.047108058035714294
e:0.021428571428571432
f:0.039642857142857146
g:0.03266071428571429

Than, i've tried next graph:
>a 3
b
c
d

but result is still about 0.2.

Graph like:
>a 1
b
b 1
c
c 1
a

Work perfectly and give a sum ~1
Also, I tried to generate RandomGraph and run PageRank, total sum of ranks about 1 as well. But with real graph it just not working (or may be i'm doing something wrong).

Could you please assist me?

bmckown:
I wrote the implementation for pagerank.  @pankajgupta I will take a look at this.

alexloginov:
Note: I've tried to override GraphReader and put StoredGraphDir.BothInOut (assumed, that problem could be in missing InboundNodes references), but it didn't help.

bmckown:
A quick note:  release 5.1.0 has a new implementation of all link analysis algorithms including PageRank and Hits.  However, when running the above graph on the new version, there is a divide by zero error occurring yielding infinity for nodes with no outbound neighbors.  Good catch @alexloginov.  I will fix the implementation and submit a patch.

[...]

PR #187, opened 2015-06-17
Modify page rank to work with dangling nodes

bmckown:
Apparently we did not handle dangling nodes at all in the implementation of page rank.  This new implementation taken from [here] (https://github.com/networkx/networkx/blob/master/networkx/algorithms/link_analysis/pagerank_alg.py) correctly handles dangling nodes.

This PR directly addresses #186 

cc @pankajgupta 
cc @alexloginov

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
@@ -15,6 +15,7 @@ package com.twitter.cassovary.algorithms.linkanalysis
 
 import com.twitter.cassovary.graph.{DirectedGraph, Node}
 import com.twitter.cassovary.util.Progress
+import scala.collection.immutable.BitSet
 
 /**
  * Parameters for PageRank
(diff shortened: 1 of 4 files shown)


alexloginov:
@bmckown 
It looks promising!
BTW, what way is preferred: use dependency from maven or download jar from github releases?

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
```diff
@@ -44,15 +44,28 @@ case class PageRankIterationState(pageRank: Array[Double], error: Double, iterat
 class PageRank(graph: DirectedGraph[Node], params: PageRankParams = PageRankParams())
   extends LinkAnalysis[PageRankIterationState](graph, params, "pagerank") {
 
-  private val outboundCount = new Array[Double](graph.maxNodeId + 1)
+  private lazy val outboundCount = new Array[Double](graph.maxNodeId + 1)
+  private val nodeCount = graph.nodeCount
 
-  graph foreach { node => efficientNeighbors(node) foreach { nbr => outboundCount(nbr) += 1} }
+  private val dangling = if (isInStored) {
```
pankajgupta: suggest changing 'dangling' to 'zeroOutgoing' to be clearer.

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
```diff
@@ -44,15 +44,28 @@ case class PageRankIterationState(pageRank: Array[Double], error: Double, iterat
...
+      neighbors foreach { nbr => outboundCount(nbr) += 1 }
+      partialSet ++ neighbors.toSet
+    }
+    graph.map { _.id }.toSet -- nonDangling
+  }
+  else {
+    graph.foldLeft(Set.empty[Int]){ (partialSet, node) =>
+      if (efficientNeighbors(node).isEmpty) partialSet + node.id else partialSet
+    }
+  }
```
pankajgupta: don't think you need to invoke if (isInStored) here. This is just:
graph.toSet.filter(outBoundCount(_) == 0)

cassovary-core/src/test/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRankSpec.scala
```diff
@@ -91,7 +91,17 @@ class PageRankSpec extends WordSpec with Matchers {
...
+
+    "Handle dangling nodes correctly when graph is stored only out" in {
+      val target = Array(0.0, 0.11622, 0.21500, 0.16763, 0.16763, 0.07625, 0.14106, 0.11622)
+      testGraphHelper(TestGraphs.dangling_g7, target)
+    }
+
+    "Handle dangling nodes correctly when graph is stored only in" in {
+      val target = Array(0.0, 0.12085, 0.15194, 0.05628, 0.05628, 0.28697, 0.27141, 0.05628)
+      testGraphHelper(TestGraphs.dangling_g7_in, target)
     }
```
pankajgupta: can you have a very small say 3-node graph A->C, B->C (i.e., node B is zeroOutgoing) to verify the exact pagerank values in the test?

pankajgupta:
The two are in sync -- but using dependency from maven is always preferred.


On Wed, Jun 17, 2015 at 2:08 PM, Alexander Loginov <notifications@github.com

cassovary-core/src/main/scala/com/twitter/cassovary/algorithms/linkanalysis/PageRank.scala
bmckown: I have a question--this could be a design choice or a bug and want to know which one it is.  Lets take `TestGraphs.dangling_g7_in` as an example which is a graph with man zeroOutgoing nodes and only stores the in bound direction.  NodeId 2 has two inbound connections (3, 4) and itself forms an inbound connection to (1, 6).  In my mind, the following code should pass a test:

```scala
   val graph = TestGraphs.dangling_g7_in
   val node2 = graph.getNodeById(2).get
   node2.inboundCount shouldEqual 2
   node2.outboundCount shouldEqual 2
```
However, the code in its current state returns `node2.outboundCount` as being 0 since it does not store that connection.

Was this chosen to prevent an expensive computation from occuring?  Obviously to get the outbound connections for a graph stored OnlyIn would be much more costly because one could potentially parse every node looking for incident connection.  If you could clear this up for me that would be much appreciated.

pankajgupta: This is correct. outboundCount is 0 because there are no outboundnodes in the graph that only stores the in direction. For a graph which _only_ stores the 'in' direction (i.e., not the out direction as well), we would need to calculate explicitly outboundCount.

bmckown: So that is why I check isInStored here.  Basically, I don't need to build the outboundCount array if the graph is out stored or both stored since that attribute is stored at the node level.  I only need to build that array if it is in stored (which is an expensive calculation because we need to visit each node).  By checking isInStored here, we can save on the calculation for outboundNodes in the determination of dangling nodes.

pankajgupta:
1. Let's use a BitSet explicitly. We are anyway indexing array by id. Default Set seems to be HashSet and I am concerned it might blow up in memory.
2. Please add some one line comments for readability.

Looks good otherwise.

bmckown:
Updated and squashed.

Issue #188, opened 2015-06-23
Exception loading graph in SharedArrayBasedDirectedGraph


plofgren:
Today when I was loading a simple graph into SharedArrayBasedDirectedGraph, I received an ArrayIndexOutOfBoundsException.  The graph is [here](https://github.com/plofgren/cassovary/blob/load-bug/cassovary-core/src/test/resources/graphs/test_graph.txt) and a minimal main class demonstrating the error is [here](https://github.com/plofgren/cassovary/blob/load-bug/cassovary-core/src/main/scala/LoadBugMain.scala).  The stack trace is below.  I just pulled the current version of cassovary, and my scala version is 2.11.6.
When I change `reader.toSharedArrayBasedDirectedGraph` to `reader.toArrayBasedDirectedGraph` the graph loads correctly.  I'll switch to using ArrayBasedDirectedGraph, but I wanted to report this in case someone can see what causes this and help others avoid this.

Thanks!

Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
	at com.twitter.cassovary.graph.SharedArrayBasedDirectedGraph$SharedArrayBasedDirectedGraphCo$$$$308510d7b33aba1419de2daaa79926$$$$nonfun$8$$anonfun$apply$15$$anonfun$apply$16$$anonfun$apply$6.apply$mcVI$sp(SharedArrayBasedDirectedGraph.scala:278)
	at com.twitter.cassovary.graph.SharedArrayBasedDirectedGraph$SharedArrayBasedDirectedGraphCo$$$$308510d7b33aba1419de2daaa79926$$$$nonfun$8$$anonfun$apply$15$$anonfun$apply$16$$anonfun$apply$6.apply(SharedArrayBasedDirectedGraph.scala:276)
	at com.twitter.cassovary.graph.SharedArrayBasedDirectedGraph$SharedArrayBasedDirectedGraphCo$$$$308510d7b33aba1419de2daaa79926$$$$nonfun$8$$anonfun$apply$15$$anonfun$apply$16$$anonfun$apply$6.apply(SharedArrayBasedDirectedGraph.scala:276)
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofInt.foreach(ArrayOps.scala:234)
[...]

PR #189, opened 2015-06-26
Issue #188

szymonm:
This should solve #188


cassovary-examples/src/main/scala/HelloLoadSharedGraph.scala
```diff
@@ -0,0 +1,38 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+import com.twitter.cassovary.graph.StoredGraphDir
+import com.twitter.cassovary.graph.StoredGraphDir.StoredGraphDir
+import com.twitter.cassovary.util.NodeNumberer
+import com.twitter.cassovary.util.io.AdjacencyListGraphReader
+
+object HelloLoadSharedGraph {
```
pankajgupta: instead of adding this file as well as a new graph, can we add a test?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/SharedArrayBasedDirectedGraph.scala
```diff
@@ -242,10 +242,9 @@ object SharedArrayBasedDirectedGraph {
         statsReceiver.timeFuture("graph_load_fill_in_edge_lengths_and_offsets") {
           val futures = (0 until numOfShards).map {
             shard => futurePool {
-              var id = 0
-              (0 until (maxNodeId + 1) / numOfShards).foreach {
-                nodeInShard =>
-                  id = nodeInShard * numOfShards + shard
+              // we assume modulo hashing below
+              (shard until (maxNodeId + 1) by numOfShards).foreach {
```
pankajgupta: ooh -- the bug was wrong loop variables... nice catch.

cassovary-examples/src/main/scala/HelloLoadSharedGraph.scala
szymonm: Added test, maybe we can leave this file as an example if we already have it?

Issue #185, opened 2015-05-28
ListOfEdgesGraphReader: Handle duplicate edges and/or maintain specified/sorted order


pankajgupta:
com.twitter.cassovary.util.io.ListOfEdgesGraphReader reads each edge in input one by one and will form an adjacency list (neighbor nodes of a source node) in the same order as it sees it and does not handle duplicates. Add variables to this class constructor to:
1. Handle duplicate edges or not
2. Keep original order in input or keep neighbors of a node in sorted order

https://github.com/twitter/cassovary/pull/184 (later reverted) only handled duplicate edges but in doing so did not keep the original order of edges seen.

tlasica:
I see two approaches possible:
a) using smart structure instead of ArrayBuffer[Int] to do dedup during adding neighbours
b) do post processing for sorting or duplicate elimination

I assume that the signature of below function should remain unchanged to makes things easier
```scala
def readEdgesBySource(): (Int2ObjectMap[ArrayBuffer[Int]], Int2IntArrayMap) = {
```

Solution (a) introduces additional memory overhead for keeping both set and list in case of unsorted with duplicates elimination plus some extra memory copying at the end of the processing. It should come in zero cost if neither deduplication nor sorting is applied. 

Solution (b) can do the postprocessing on each value in the edgesBySource map. It will cost additional memory per neighbours list (much less than solution (a)). Memory copying will happen also in this case for some cases e.g. dedup with original order.

Time complexity ibn both cases is similar of n * O(klogk) for sorting (when n is number of nodes and k is #of neighbours). Duplicate elimination in both cases is feasible in additional O(m) cost (m = #edges).

Do you have any suggestions which approach to follow?
I have (a) implemented and tested - as it seemed to be more elegant one :-)

pankajgupta:
IIUC you are saying that (a) requires at least 2x memory in your implementation as you would keep both a set and a list? If so, that is just too much. You could keep only a SortedSet in (a) and then convert to Array at the end. Alternatively, post-processing as in (b) looks good too.

tlasica:
Yes, in solution (a) there is twice as much memory used. Unfortunately it is impossible to use only  SortedSet(a) if you want to keep the order of neighbours as given in the file. It is possible if we decide that deduplication=true and sort=false is not acceptable.

pankajgupta:
Makes sense -- (b) sounds more flexible then, doesn't it?

PR #192, opened 2015-07-16
Implementation of #185: optional sorting and deduplication of adj edges

tlasica:
This pull request implements issue #185 

I assumed that reader is used to traverse the file only once, so the best place to do sorting or duplicate elimination is next() when already a memory copying *toArray* happen. I made two vals private to make sure it is accessed via iterator only.

I have added two new constructor parameters with default values to keep backward compatibility, but other option may be to add one parameter as trait and do some stacking.
```scala
trait EdgeListProcessor
trait EdgeListSort extends EdgeListProcessor
trait EdgeListDuplicateElimination extends EdgeListProcessor
```

I was also a bit surprised by the fact that in Spec after adding a test file *toy_6nodes_list_StringIds_dup_edges.txt* it was used instead of *toy_6nodes_list_StringIds.txt* making tests failed. I changed to full names with extention to ommit such problems in future.

One more topic: is it is indeed more efficient to keep two maps: one for edges and one for max out node id? Making it one is quite simple, will save some memory and will save some map lookups in graphs with many edges. But I may be not aware of some other requirements or implementation details.


cassovary-core/src/test/resources/graphs/toy_6nodes_list_StringIds_unsorted.txt
```diff
@@ -0,0 +1,9 @@
+# 6 nodes cycle with String ids and unsorted a-{c,b}, {c->d,b}
```
pankajgupta: is "a-{c..." this a minor typo and you intended "a->{c..."

cassovary-core/src/test/resources/graphs/toy_6nodes_list_StringIds_unsorted_dup_edges.txt
```diff
@@ -0,0 +1,15 @@
+# 6 nodes cycle with String ids and unsorted a-{c,b}, {c->d,b}
```
pankajgupta: ditto

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
```diff
@@ -57,7 +57,9 @@ class ListOfEdgesGraphReader[T](
   val directory: String,
   override val prefixFileNames: String,
   val nodeNumberer: NodeNumberer[T],
-  idReader: (String => T)
+  idReader: (String => T),
+  removeDuplicates: Boolean = false,
+  sortNeighbours: Boolean = false
```
pankajgupta: prefer the spelling sortNeighbors to be consistent with the rest of the code

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
```diff
@@ -57,7 +57,9 @@ class ListOfEdgesGraphReader[T](
   val directory: String,
   override val prefixFileNames: String,
   val nodeNumberer: NodeNumberer[T],
-  idReader: (String => T)
+  idReader: (String => T),
```
pankajgupta: please update comments as well

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
```diff
@@ -72,6 +74,7 @@ class ListOfEdgesGraphReader[T](
 
       var lastLineParsed = 0
 
+
```
pankajgupta: delete extra line

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
```diff
@@ -83,6 +86,7 @@ class ListOfEdgesGraphReader[T](
         val nodeMaxOutEdgeId = new Int2IntArrayMap()
 
         def updateNodeMaxOutEdgeId(node: Int, out: Int) {
+          //TODO: make it getOrDefault + put: simple and less ops
```
pankajgupta: remove the TODO -- either leave it like this or change it

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
```diff
@@ -107,21 +111,32 @@ class ListOfEdgesGraphReader[T](
         }
         log.info("Finished reading from file %s...\n", filename)
         Source.fromFile(filename).close()
+
```
pankajgupta: delete extra line

cassovary-core/src/test/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReaderSpec.scala
```diff
@@ -91,7 +125,76 @@ class ListOfEdgesGraphReaderSpec extends WordSpec with GraphBehaviours[Node] wit
...
+      }
+
+      "provide the correct graph properties with duplicate elimination reader" in {
+        new GraphWithStringIdsWithDedupReader {
+          graph.nodeCount should be(6)
+          graph.edgeCount should be(6L)
+          graph.maxNodeId should be(5)
+        }
+      }
+      "contain the right nodes and edges with duplicate elimination reader" in {
```
pankajgupta: add extra line

cassovary-core/src/test/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReaderSpec.scala
```diff
@@ -53,6 +56,37 @@ class ListOfEdgesGraphReaderSpec extends WordSpec with GraphBehaviours[Node] wit
       idReader = _.toLong).toSharedArrayBasedDirectedGraph()
   }
 
+  trait GraphWithStringIdsWithSortReader {
```
pankajgupta: there is a lot of duplication of code here because of the various cases to cover. Would be better to not instantiate named traits but classes with different variables e.g., class Graph(fileName: String, sortNeighbors, removeDuplicates) and then do the same tests for all the combinations of the variables sortNeighbors and removeDuplicates

pankajgupta:
Your implementation logic of post-processing looks good.

Let's leave the two maps around -- good for readability.

cassovary-core/src/test/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReaderSpec.scala
tlasica: That's right. I am aware of it it but wanted to make sure the approach is correct. I will keep separate tests but reduce traits.

tlasica:
The deduplication of tests code can be done in many ways. It could be done even further by generation of (dup,sort) combinations in one test. In my opinion it is too much as it makes tracking failing scenarios more difficult.

pankajgupta:
Yes, there is a balance for sure. Will let you decide what is the right tradeoff of readability/reusability

cassovary-core/src/test/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReaderSpec.scala
```diff
@@ -79,19 +82,95 @@ class ListOfEdgesGraphReaderSpec extends WordSpec with GraphBehaviours[Node] wit
...
+          behave like renumberedGraphEquivalentToMap(graph, stringGraphMap, seqNumberer)
+        }
+      }
+
+      "on file with duplicated edges provide the correct graph properties with duplicate elimination reader" in {
+        new GraphWithStringIds("toy_6nodes_list_StringIds_dup_edges.txt", removeDup = true) {
+          hasCorrectProperties(graph)
+        }
+      }
+      "on file with duplicated edges contain the right nodes and edges with duplicate elimination reader" in {
```
pankajgupta: minor: please be consistent regarding empty lines

pankajgupta:
Except for the minot comment about line spacing, looks good to me -- please do a squash commit and then I will merge

cassovary-core/src/test/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReaderSpec.scala
tlasica: Frankly speaking I tried :-) I kept one empty line when the graph/reader changes and no empty lines for different tests of the same graph.

pankajgupta:
ok, i will merge after your squash commit

tlasica:
Is it fine to create a new pull request with squashed commit (it seems it is not possible to push to remote after locally squashing commits)?

AnishShah:
@tlasica Did you try `git push -f` ?

tlasica:
Thanks for hint, indeed it works with -f.

PR #194, opened 2015-07-19
Remove the concept of internal/external id in Bipartite Graph

AnishShah:


cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -49,11 +49,12 @@ trait BipartiteNode extends Node {
  *        will be internalized to nodeId * -1
  * @param inboundNodes original (positive) ids of the nodes on the RHS pointed by in-coming edges
  * @param outboundNodes original (positive) ids of the nodes on the RHS pointed by out-going edges
+ * @param unique whether node id is unique in the whole graph
  */
 class LeftNode(nodeId: Int, val inboundNodes: Seq[Int],
-    val outboundNodes: Seq[Int]) extends BipartiteNode {
+    val outboundNodes: Seq[Int], val unique: Boolean = false) extends BipartiteNode {
```
pankajgupta: unique should be a property of the graph, not each node

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -115,7 +118,8 @@ case class BipartiteSide(nodes: Array[BipartiteNode], numOfNodes: Int, numOfOutE
 class BipartiteGraph(val leftNodes: Array[BipartiteNode], val leftNodeCount: Int,
                      val leftOutEdgeCount: Long, val rightNodes: Array[BipartiteNode],
                      val rightNodeCount: Int, val rightOutEdgeCount: Long,
-                     val bipartiteGraphDir: BipartiteGraphDir.BipartiteGraphDir) extends Graph[BipartiteNode] {
+                     val bipartiteGraphDir: BipartiteGraphDir.BipartiteGraphDir,
+                     val isLeftNode: (Int => Option[Boolean]) = (_ => None)) extends Graph[BipartiteNode] {
```
pankajgupta: if leftNode and rightNodes are explicitly specified, the function isLeftNode is not useful to also specify. So I would not have this extra isLeftNode parameter.

Adding 'unique' here in the constructor of the graph makes sense -- let's call it nodeIdsUnique (please add a comment about what this means).

pankajgupta: This also means that is unique is false, new leftNodes and rightNodes will need to be constructed with the correct ids. I don't see a way around that.

AnishShah:
@pankajgupta I have made some changes. Please have a look. 
If node IDs are not unique, then I'm negating the IDs of Left Nodes.
I don't think this will increase the time complexity as this was done at Node Level before instead inside the Bipartite Graph and memory should also be same. What do you think?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
```diff
@@ -278,6 +278,42 @@ object TestGraphs {
     new BipartiteGraph(leftSide, rightSide, BipartiteGraphDir.Both)
   }
 
+  def bipartiteGraphWithUniqueNodeIds = {
+
+    val leftNodes = new Array[BipartiteNode](7)
+
+    var inBounds = Array(1, 3, 5);
```
pankajgupta: consider not using vars. In fact no point using intermediate variables inBounds and outBounds. Just directly use as in new LeftNode(2, Array(...), Array(...))

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -61,27 +60,13 @@ class LeftNode(nodeId: Int, val inboundNodes: Seq[Int],
  * all of its in and out edges point to nodes in the LHS, and thus all edge ids have values
  * of the real id * -1
  * @param id the id of the node (unique on the RHS)
- * @param in original (negative) ids of the nodes on the LHS pointed by in-coming edges
- * @param out original (negative) ids of the nodes on the LHS pointed by out-going edges
+ * @param inboundNodes ids of the nodes on the LHS pointed by in-coming edges
+ * @param outboundNodes ids of the nodes on the LHS pointed by out-going edges
+ * inboundNodes and outboundNodes IDs should be positive if all the node IDS are unique in a graph
```
pankajgupta: s/IDS/ids

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -110,22 +95,52 @@ case class BipartiteSide(nodes: Array[BipartiteNode], numOfNodes: Int, numOfOutE
  * semantic meaning between the left and right sides; but if the graph is of direction
  * BipartiteBoth, then it could carry two semantic
  * meanings in one graph (e.g. list-to-user membership and user-to-list followship).
- *
+ * @param nodeIdsUnique true if node Ids are unique in the whole graph
+ *                      and false if unique only on each side (LHS and RHS)
+ * If the node IDs are unique only on each side, then the Left Nodes IDs are negated,
```
pankajgupta: try to be consistent -- sometimes it is "IDS", "Ids", "ids" etc.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -110,22 +95,52 @@ case class BipartiteSide(nodes: Array[BipartiteNode], numOfNodes: Int, numOfOutE
...
-class BipartiteGraph(val leftNodes: Array[BipartiteNode], val leftNodeCount: Int,
-                     val leftOutEdgeCount: Long, val rightNodes: Array[BipartiteNode],
+class BipartiteGraph(leftBipartiteNodes: Array[BipartiteNode], val leftNodeCount: Int,
+                     val leftOutEdgeCount: Long, rightBipartiteNodes: Array[BipartiteNode],
                      val rightNodeCount: Int, val rightOutEdgeCount: Long,
-                     val bipartiteGraphDir: BipartiteGraphDir.BipartiteGraphDir) extends Graph[BipartiteNode] {
+                     val bipartiteGraphDir: BipartiteGraphDir.BipartiteGraphDir,
+                     val nodeIdsUnique: Boolean = false) extends Graph[BipartiteNode] {
 
   def this(leftSide: BipartiteSide, rightSide: BipartiteSide,
```
pankajgupta: delete this convenience constructor once you have added the next constructor

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -153,18 +168,24 @@ class BipartiteGraph(val leftNodes: Array[BipartiteNode], val leftNodeCount: Int
...
     if (id == 0) {
       return None // id 0 is not allowed in bipartite graph
-    } else if (isLeftNodeId) {
-      //invert nodeId to index
-      val leftOriginalId = - id
-      if (leftOriginalId < leftNodes.length) {
-        if (leftNodes(leftOriginalId) != null) Some(leftNodes(leftOriginalId))
-        else None
-      } else None
+    } else if (!nodeIdsUnique) {
```
pankajgupta: this should be before any other code here, as id==0 is also possible in the other case

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -153,18 +168,24 @@ class BipartiteGraph(val leftNodes: Array[BipartiteNode], val leftNodeCount: Int
...
+      if (isLeftNodeId) {
+        //invert nodeId to index
+        val leftOriginalId = -id
+        if (leftOriginalId < leftNodes.length) {
```
pankajgupta: refactor this code segment into a convenience method, as the same code is used 4 times below.

if (id < arr.length && arr(id) != null) Some(arr(id)) else None

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -110,22 +95,52 @@ case class BipartiteSide(nodes: Array[BipartiteNode], numOfNodes: Int, numOfOutE
...
+           bipartiteGraphDir: BipartiteGraphDir.BipartiteGraphDir,
+           nodeIdsUnique: Boolean) =
+     this(leftSide.nodes, leftSide.numOfNodes, leftSide.numOfOutEdges,
+       rightSide.nodes, rightSide.numOfNodes, rightSide.numOfOutEdges,
+       bipartiteGraphDir, nodeIdsUnique)
 
   private val log = Logger.get
   val storedGraphDir =  StoredGraphDir.Bipartite
 
+  val leftNodes = {
```
pankajgupta: creation of new left and right nodes is actually pretty expensive in both space and time. Space is doubled as two new arrays have been created, also each edge is being touched to negate the inbound and outbound nodes... So, I'd suggest adding a comment that if efficiency is desired, it is best to ensure that nodeIdsUnique=true and in fact I'd make it the default.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
AnishShah: Do you mean id can be 0 when nodes are unique in the whole graph?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/TestGraph.scala
AnishShah: Will do this for previous test graph too.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
AnishShah: I agree that nodeIdsUnique should be true & it should be the default value.

But, my intention of keeping default value to false was that it doesn't break the existing code even after upgrading to this new version of cassovary. Should we remove the default value completely?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
AnishShah: Scala allows only one constructor to have default values. So, we need both these constructors so that it doesn't break existing tests. Again, should we remove the default value completely as I said below and change tests?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
pankajgupta: Let's keep the default, but change it to true. We will upgrade the library version appropriately on next publish.

AnishShah:
@pankajgupta  Please have a look. I have made the changes..

pankajgupta:
Looks good. Please put in a squash commit and I will merge.

AnishShah:
Done :+1:

PR #195, opened 2015-07-21
new sbt

zjpjack:
Add PermSize option to avoid out of memory problem.


bmckown:
Should we bump to 13.8 while we are updating the build?  @pankajgupta

pankajgupta:
@bmckown Sure, but separate concern so deserves a separate pull request.

PR #198, opened 2015-08-26
Implement efficient dynamic graph for the case of concurrent reads an…

plofgren:
…d single-threaded writes.
@ashishgpersonal When you have time, could you code review this?
Also, two questions:
1. Currently addEdges does not check for duplicate edges, because e.g. if Obama gets a new in-neighbor I don't want to scan all his current in-neighbors, and I don't want the overhead of using hashMaps rather than arrays.  I think duplicate edges should be rare enough for us to ignore for now--is that OK?  It's not hard to add the check, it's just an efficiency question.
2. I don't think it's easy to test race conditions without inserting artificial `Thread.sleep` calls.  Could you look carefully at the `append` method in `SemiSynchronizedIntArrayList` and see if you think it's thread safe with concurrent readers?

Thanks


plofgren:
(I meant to pull request this in to plofgren/cassovary first. Closing for now on twitter/cassovary.)

PR #201, opened 2015-09-12
Memory mapped graph

plofgren:
This is a graph that stores its data in a memory mapped file. This enables two nice features:
1) No object overhead.  The memory used for n nodes and m edges with both in-neighbor and out-neighbor access is exactly 8 + 16*n + 8*m bytes
2) Nearly instant graph loading.  The twitter-2010 graph with 1.5 billion edges loads in seconds if the binary graph file is already in OS cache, and otherwise loads in the time it takes the OS to transfer the file from disk to RAM.

@pankajgupta

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
@@ -1,0 +1,126 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.graph
+
+import com.twitter.cassovary.util.io.{IntLongSource, MemoryMappedIntLongSource}
+import java.io._
+
+/**
+ * A graph which reads edge data from a memory mapped file.  There is no object overhead per
+ * node: the memory used for n nodes and m edges with both in-neighbor and out-neighbor access is
+ * exactly 16 + 16*n + 8*m bytes. Also, loading is very fast because no parsing of text is required.
+ * Loading time is exactly the time it takes the operating system to map data from disk into
+ * memory.  Nodes are numbered sequentially from 0 to nodeCount - 1 and must be a range of this
+ * form (i.e. nodeCount == maxNodeId + 1).
+ *
+ * When transforming a graph where nodeCount <= maxNodeId
+ * to this format, new nodes with no neighbors will be implicitly created.  Currently only supports
+ * storing both in-neighbors and out-neighbors of nodes (StoredGraphDir.BothInOut). The binary
+ * format is currently subject to change.  Node objects are created on demand when getNodeById is
+ * called.
+ */
+
+/* Storage format
+byteCount  data
+8          (reserved, later use for versioning or indicating undirected vs directed)
+8          n (i. e. the number of nodes).  Currently must be less than 2^31.
+8*(n+1)    Offsets into out-neighbor data. Index i (a Long) points to the out-neighbor data of
+           node i.  The out-neighbor data must be stored in sequential order by id, as the
+           outegree of node i is computed from the difference in offset between node i+1 and node i.
+           Index n is needed to compute the outdegree of node n - 1.
+8*(n+1)    Offsets into in-neighbor data (Longs) (Same interpretation as out-neighbor offsets)
+m          out-neighbor data
+m          in-neighbor data
+ */
+class MemoryMappedDirectedGraph(file: File) extends DirectedGraph[Node] {
+  val data: IntLongSource = new MemoryMappedIntLongSource(file)
+
+  val nodeCount = data.getLong(8).toInt // In the future we may want to support Long ids, so
+  // store nodeCount as Long
+
+  private def outboundOffset(id: Int): Long = data.getLong(16L + 8L * id)
+
+  private def outDegree(id: Int): Int = ((outboundOffset(id + 1) - outboundOffset(id)) / 4).toInt
+
+  private def inboundOffset(id: Int): Long = data.getLong(16L + 8L * (nodeCount + 1) + 8L * id)
+
+  private def inDegree(id: Int): Int = ((inboundOffset(id + 1) - inboundOffset(id)) / 4).toInt
+
+  /* Only created when needed (there is no array of these stored). */
+  private class MemoryMappedDirectedNode(override val id: Int) extends Node {
+    val nodeOutboundOffset = outboundOffset(id)
+    val nodeInboundOffset = inboundOffset(id)
+    def outboundNodes(): Seq[Int] = new IndexedSeq[Int] {
+      val length: Int = outDegree(id)
+      def apply(i: Int): Int =  data.getInt(nodeOutboundOffset + 4L * i)
+    }
+    def inboundNodes(): Seq[Int] = new IndexedSeq[Int] {
+      val length: Int = inDegree(id)
+      def apply(i: Int): Int =  data.getInt(nodeInboundOffset + 4L * i)
+    }
+  }
+
+  def getNodeById(id: Int): Option[Node] =
+    if(0 <= id && id < nodeCount) {
+      Some(new MemoryMappedDirectedNode(id))
+    } else {
+      None
+    }
+
+  def iterator: Iterator[Node] = (0 to nodeCount).iterator flatMap (i => getNodeById(i))
+
+  lazy val edgeCount: Long = outboundOffset(nodeCount) - outboundOffset(0)
+
+  override lazy val maxNodeId = nodeCount - 1
+
+  val storedGraphDir = StoredGraphDir.BothInOut
+}
+
+object MemoryMappedDirectedGraph {
+  /** Writes the given graph to the given file (overwriting it if it exists) in the current binary
+   * format.
+   */
+  def graphToFile(graph: DirectedGraph[Node], file: File): Unit = {
+    val n = graph.maxNodeId + 1 // includes both 0 and maxNodeId as ids
+    val out = new DataOutputStream(new BufferedOutputStream(new FileOutputStream(file)))
+    out.writeLong(0)
+    out.writeLong(n)
+    //The outneighbor data starts after the initial 8 bytes, n+1 Longs for outneighbors, and n+1
+    // Longs for in-neighbors
+    var outboundOffset = 16L + 8L * (n + 1) * 2
+    for (i <- 0 until n) {
+      out.writeLong(outboundOffset)
+      outboundOffset += 4 * (graph.getNodeById(i) map (_.outboundCount)).getOrElse(0)
+    }
+    out.writeLong(outboundOffset) // Needed to compute outdegree of node n-1
+
+    // The inbound data starts immediately after the outbound data
+    var inboundOffset = outboundOffset
+    for (i <- 0 until n) {
+      out.writeLong(inboundOffset)
+      inboundOffset += 4 * (graph.getNodeById(i) map (_.inboundCount)).getOrElse(0)
+    }
+    out.writeLong(inboundOffset) // Needed to compute indegree of node n-1
+
+    for (i <- 0 until n) {
+      for (v <- (graph.getNodeById(i) map (_.outboundNodes())).getOrElse(Nil)) {
+        out.writeInt(v)
+      }
+    }
+    for (i <- 0 until n) {
+      for (v <- (graph.getNodeById(i) map (_.inboundNodes())).getOrElse(Nil)) {
+        out.writeInt(v)
+      }
+    }
+    out.close()
+  }
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2015 Teapot, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.util.io
+
+import java.io.File
+import java.nio.channels.FileChannel
+import java.nio.channels.FileChannel.MapMode
+import java.nio.file.StandardOpenOption
+import java.nio.MappedByteBuffer
+
+/**
+ * Represents an arbitrarily large sequence of bytes which can be interpreted as ints or longs.
+ */
+trait IntLongSource {
+  /** Returns the int at the given byte index, which must be a multiple of 4. */
+  def getInt(index: Long): Int
+  /** Returns the Long at the given byte index, which must be a multiple of 8. */
+  def getLong(index: Long): Long
+}
+
+/**
+ * Wraps a sequence of FileChannels to enable random access on a memory mapped file of arbitrary size.
+ * Motivation: FileChannel.open only supports 2GB at a time.
+ * @param file The file containing binary data.
+ */
+
+class MemoryMappedIntLongSource(file: File) extends IntLongSource {
+  def this(fileName: String) = this(new File(fileName))
+
+  private val fileChannel = FileChannel.open(file.toPath, StandardOpenOption.READ)
+
+  // Each buffer uses int addressing, so can only access 2GB.  We'll use multiple buffers,
+  // and access 2^30 bytes per buffer.
+  // We use 2^30 bytes per buffer rather than 2^31 because fileChannel.map doesn't currently
+  // support size 2^31.
+  private val bytesPerBuffer = 1L << 30
+  // ceiling of file.length() / bytesPerBuffer
+  private val bufferCount = ((file.length() + bytesPerBuffer - 1) / bytesPerBuffer).toInt
+  private val byteBuffers: Array[MappedByteBuffer] = (0 until bufferCount).toArray map { bufferIndex =>
+    val size = if (bufferIndex + 1 < bufferCount)
+      bytesPerBuffer
+    else
+      file.length - (bufferCount - 1L) * bytesPerBuffer
+    fileChannel.map(MapMode.READ_ONLY, bufferIndex * bytesPerBuffer, size)
+  }
+  private def bufferIndex(index: Long): Int = (index >> 30).toInt
+  private def indexWithinBuffer(index: Long): Int = (index & 0x3FFFFFFF).toInt
+
+  def getInt(index: Long): Int = byteBuffers(bufferIndex(index)).getInt(indexWithinBuffer(index))
+
+  def getLong(index: Long): Long = byteBuffers(bufferIndex(index)).getLong(indexWithinBuffer(index))
+}
--- a/cassovary-core/src/test/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraphSpec.scala
+++ b/cassovary-core/src/test/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraphSpec.scala
@@ -1,0 +1,39 @@
+package com.twitter.cassovary.graph
+
+import java.io.File
+import java.nio.file.NoSuchFileException
+
+import org.scalatest.{Matchers, WordSpec}
+
+class MemoryMappedDirectedGraphSpec extends WordSpec with Matchers {
+  val testGraph1 = ArrayBasedDirectedGraph.apply(
+    Iterable(
+      NodeIdEdgesMaxId(1, Array(2, 3)),
+      NodeIdEdgesMaxId(3, Array(1, 2)),
+      NodeIdEdgesMaxId(5, Array(1))
+    ),
+    StoredGraphDir.BothInOut,
+    NeighborsSortingStrategy.LeaveUnsorted)
+
+  "A MemoryMappedDirectedGraph" should {
+    "correctly store and read a graph" in {
+      val tempFile = File.createTempFile("graph1", ".bin")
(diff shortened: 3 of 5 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
+package com.twitter.cassovary.graph
```
plofgren: I should add a copyright statement.

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
```diff
@@ -0,0 +1,48 @@
...
+import java.io.File
+import java.nio.MappedByteBuffer
+import java.nio.channels.FileChannel
+import java.nio.channels.FileChannel.MapMode
+import java.nio.file.StandardOpenOption
+
+/**
+ * Represents an arbitrarily large array of bytes which can be interpreted as ints or longs.
+ */
+trait IntLongSource {
```
plofgren: I should probably rename this, but I'm not sure to what.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+
+import java.io.{BufferedOutputStream, DataOutputStream, File, FileOutputStream}
```
pankajgupta: style comment: 4 or more imports -- just use java.io._

pankajgupta: also imports in alphabetical order

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+
+import com.twitter.cassovary.graph.StoredGraphDir.StoredGraphDir
+import com.twitter.cassovary.util.NodeNumberer
+import com.twitter.cassovary.util.io.{AdjacencyListGraphReader, IntLongSource, MemoryMappedIntLongSource}
+
+/**
+ * A graph which reads edge data from a memory mapped file.  There is no object overhead per node: the memory
+ * used for n nodes and m edges with both in-neighbor and out-neighbor access is exactly 8 + 16*n + 8*m bytes.
+ * Also, Loading is very fast because no parsing of text is required.  Loading time is exactly the time it takes the operating system
```
pankajgupta: some lines are too long, like this one. Might be good to limit the line size to 100

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
+package com.twitter.cassovary.graph
+
+import java.io.{BufferedOutputStream, DataOutputStream, File, FileOutputStream}
+
+import com.twitter.cassovary.graph.StoredGraphDir.StoredGraphDir
+import com.twitter.cassovary.util.NodeNumberer
+import com.twitter.cassovary.util.io.{AdjacencyListGraphReader, IntLongSource, MemoryMappedIntLongSource}
+
+/**
+ * A graph which reads edge data from a memory mapped file.  There is no object overhead per node: the memory
```
pankajgupta: may be also say that nodes are constructed on demand

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+ * used for n nodes and m edges with both in-neighbor and out-neighbor access is exactly 8 + 16*n + 8*m bytes.
+ * Also, Loading is very fast because no parsing of text is required.  Loading time is exactly the time it takes the operating system
+ * to map data from disk into memory.
+ *
+ * Nodes are numbered sequentially from 0 to nodeCount - 1 and must be a range of this form (i.e. nodeCount == maxNodeId + 1).
+ * When transforming a graph where nodeCount <= maxNodeId to this format, new nodes with no neighbors will be implicitly created.
+ *
+ * Currently only supports storing both in-neighbors and out-neighbors of nodes (StoredGraphDir.BothInOut).
```
pankajgupta: Seems trivial to extend it to other cases (not suggesting you do it in this PR itself), or do you see any issue?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+ * The binary format is currently subject to change.
+ */
+
+/* Storage format
+byteCount  data
+4          (reserved, later use for versioning or indicating undirected vs directed)
```
pankajgupta: Suggest making both these fields as 8B long -- then your format is completely future proof.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+4          n (i. e. the number of nodes)
+8*(n+1)    Offsets into out-neighbor data. Index i (a Long) points to the out-neighbor data of node i.
+           Index n is needed to compute the outdegree of node n - 1.
+8*(n+1)    Offsets into in-neighbor data (Longs) (Same interpretation as out-neighbor offsets)
+m          out-neighbor data
```
pankajgupta: Because of the Long offsets, the out-neighbor data and in-neighbor data can be arbitrarily mixed as long as all neighbors of one direction are together for one node. You might want to point that out.

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
pankajgupta: s/array/sequence

name is fine

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
pankajgupta: Ahh, you need them together for quick degree calculation...

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
```diff
@@ -0,0 +1,48 @@
...
+ * @param file The file containing binary data.
+ */
+
+class MemoryMappedIntLongSource(file: File) extends IntLongSource {
+  def this(fileName: String) = this(new File(fileName))
+
+  private val fileChannel = FileChannel.open(file.toPath, StandardOpenOption.READ)
+
+  // Each buffer uses int addressing, so can only access 2GB.  We'll use multiple buffers,
+  // and access 2^30 bytes per buffer
```
pankajgupta: why not 2^31 ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
```diff
@@ -0,0 +1,48 @@
...
+    val size = if (bufferIndex + 1 < bufferCount)
+      bytesPerBuffer
+    else
+      file.length - (bufferCount - 1L) * bytesPerBuffer
+    fileChannel.map(MapMode.READ_ONLY, bufferIndex * bytesPerBuffer, size)
+  }
+  private def bufferIndex(index: Long): Int = (index >> 30).toInt
+  private def indexWithinBuffer(index: Long): Int = (index & 0x3FFFFFFF).toInt
+
+  override def getInt(index: Long): Int = byteBuffers(bufferIndex(index)).getInt(indexWithinBuffer(index))
```
pankajgupta: no need for 'override' keywords

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+m          in-neighbor data
+ */
+class MemoryMappedDirectedGraph(file: File) extends DirectedGraph[Node] {
+  val data: IntLongSource = new MemoryMappedIntLongSource(file)
+
+  override val nodeCount = data.getInt(4)
```
pankajgupta: don't need keyword 'override' here and below

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+  override def getNodeById(id: Int): Option[Node] =
+    if(0 <= id && id < nodeCount) {
+      Some(new MemoryMappedDirectedNode(id))
+    } else {
+      None
+    }
+
+  override def iterator: Iterator[Node] = (0 to nodeCount).iterator flatMap (i => getNodeById(i))
+
+  override def edgeCount: Long = outboundOffset(nodeCount) - outboundOffset(0)
```
pankajgupta: s/def/lazy val/ -- then it will be computed once and reused

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  // Uncomment in future if maxNodeId becomes a method:
```
pankajgupta: consider just deleting this method for now

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+    for (i <- 0 until n) {
+      for (v <- (graph.getNodeById(i) map (_.inboundNodes())).getOrElse(Nil)) {
+        out.writeInt(v)
+      }
+    }
+    out.close()
+  }
+}
+
+object MemoryMappedDirectedGraphBenchmark {
```
pankajgupta: this and the main() can not be in cassovary-core as it is a library. Ideally should be in cassovary-benchmarks. I suggest you move to cassovary-examples for now and I can refactor it later.

pankajgupta:
Perhaps @szymonm might also be interested in this PR.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+ * Currently only supports storing both in-neighbors and out-neighbors of nodes (StoredGraphDir.BothInOut).
+ * The binary format is currently subject to change.
+ */
+
+/* Storage format
+byteCount  data
+4          (reserved, later use for versioning or indicating undirected vs directed)
+4          n (i. e. the number of nodes)
+8*(n+1)    Offsets into out-neighbor data. Index i (a Long) points to the out-neighbor data of node i.
+           Index n is needed to compute the outdegree of node n - 1.
```
szymonm: would be nice to specify explicitly that out-neighbor data is sorted by node id and A[i + 1] - A[i] is the outdegree.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraphSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+    Iterable(
+      NodeIdEdgesMaxId(1, Array(2, 3)),
+      NodeIdEdgesMaxId(3, Array(1, 2)),
+      NodeIdEdgesMaxId(5, Array(1))
+    ),
+    StoredGraphDir.BothInOut,
+    NeighborsSortingStrategy.LeaveUnsorted)
+
+  "A MemoryMappedDirectedGraph" should {
+    " correctly store and read a graph" in {
```
szymonm: space before correctly

cassovary-core/src/test/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraphSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+        val node = graph1.getNodeById(testNode.id).get
+        node.outboundNodes should contain theSameElementsAs (testNode.outboundNodes)
+        node.inboundNodes should contain theSameElementsAs (testNode.inboundNodes)
+      }
+      graph1.getNodeById(-1) should be(None)
+      graph1.getNodeById(6) should be(None)
+      graph1.getNodeById(1 << 29) should be(None)
+    }
+
+    " throw an error given an invalid filename" in {
```
szymonm: I believe you don't need the spaces here.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
```diff
@@ -0,0 +1,153 @@
...
+      val writeTime = (System.currentTimeMillis() - startTime) / 1000.0
+      println(s"Time to write binary graph: $writeTime")
+    } else if (args(0) == "readBin") {
+      val graph = new MemoryMappedDirectedGraph(new File(graphName))
+      println(s"outneighbors of node $testNodeId: " + graph.getNodeById(testNodeId).get.outboundNodes())
+      val loadTime = (System.currentTimeMillis() - startTime) / 1000.0
+      println(s"Time to read binary graph: $loadTime")
+    }
+  }
+}
```
szymonm: no newline

szymonm:
nice!

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
plofgren: Sounds good.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: OK, done.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Sounds good.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Sounds good.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: It should be trivial.  That's one use case I'm planning for the first 4 bytes (to indicate  if in-neighbors are stored or not).  To limit scope of this PR, I just chose both in and out neighbors, and a future PR (by me or someone else) can easily add other cases.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: You mean have both 8 bytes reserved and 8 bytes for nodeCount?  That sounds fine, but I'm curious what the motivation is.  Are you considering switching to 64 bit node ids in the future?

I'm thinking of the first 4 bytes as a format version number, so if we needed more meta data in the future we could choose a new version number, and in that format there can be more bytes for meta data.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Right.  I expanded the documentation here to state that degrees are computed based on offset i+1 minus offset i.

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
plofgren: 2^30 felt like a nicer round number :)
I was also slightly concerned about the fact that 2^31 isn't an int.
Thinking about it now, I think 2^31 would work fine, and it's slightly more efficient (half as many buffer objects), so 2^31 sounds good.

plofgren: Actually, I just tried 2^31, and FileChannel requires size at most Integer.MAX_VALUE and throws an exception given size 2^31, so we can't actually have a buffer of size 2^31.  This is a strange API decision in Java.  I'll stick with 2^30.

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
plofgren: I agree it isn't needed.  This comes up in other places, so I need to learn: what is the style guideline for override?  Should we only use it when overriding an implementation of a method? Is there a reason to not use it whenever overriding? 

This Stack overflow discussion seems somewhat against override in this case
http://stackoverflow.com/questions/5643144/using-the-override-keyword-on-implementations-of-abstract-methods
but opinions are mixed.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Sounds good.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: I don't want to forget about this, though, because right now I think calling maxNodeId will silently create millions of Node objects.  Is it OK if I leave this, and make a PR to make maxNodeId a method soon? I also need maxNodeId to be a method for the dynamic graphs. (At Teapot we've seen confusing behavior from maxNodeId on dynamic graphs).

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Sounds good

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Sounds good.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraphSpec.scala
plofgren: Sounds good.

cassovary-core/src/test/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraphSpec.scala
plofgren: Sounds good.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
plofgren: Sounds good.

plofgren:
Thanks for the feedback.  I think I've responded to everything except for a few questions I wrote comments on:
1. Motivation of 8-byte reserved area.
2. Style rule for override.  I went ahead and removed override, but I'm still curious what the rational is for removing it.
3. Keeping commented out def maxNodeId, with the intention of a follow-up PR.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
pankajgupta: Correct. 8B for nodeCount allows >2B nodes / Long node ids, something which is being worked on by @szymonm

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
pankajgupta: Interesting -- please add this finding to your comments there

pankajgupta:
Re override, simple rule I follow (but not everyone does) is to not have this keyword when not needed. Plus, the compiler will complain  in case you didn't use override but really needed to -- that way you could discover that the compiler is sequencing hierarchy of classes a certain way.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
pankajgupta: I'd suggest removing the copyright statement altogether -- that's what I am doing for any files that are new altogether.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/MemoryMappedDirectedGraph.scala
pankajgupta: Also, having things multiple of 8 is generally nice which it is in your format until the start if the out-neighbor data..

plofgren: OK, it's fine to use 8 bytes.  To fully support 8 byte nodeIds we would need to change the size of the neighbor arrays to 8 bytes per node, and probably add some compression since most of the high order bits will be 0.

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/MemoryMappedIntLongSource.scala
plofgren: Sounds good.

plofgren:
OK, I've removed the overrides, changed the reserved bytes and nodeCount to 8 bytes, and realized that I can override maxNodeId by making it a lazy val.  @pankajgupta I think I've addressed everything--is this ready to merge?
Thanks for the comments!

pankajgupta:
yep, please do a squash commit and I will merge. thanks!

PR #202, opened 2015-09-12
Efficient hashmap dynamic graph

plofgren:
A concurrent dynamic graph that stores nodes in a ConcurrentHashMap, and neighbors of each node are stored in Array[Int]s.  Only edge and node addition (not deletion) is supported.

@pankajgupta

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
@@ -1,0 +1,89 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package com.twitter.cassovary.graph
+
+import com.twitter.cassovary.graph.StoredGraphDir._
+import com.twitter.cassovary.graph.node.DynamicNode
+
+/**
+ * An efficient dynamic graph implementation which supports concurrent reading and writing.
+ * Reads happen without locks or data copying.
+ * Nodes are stored in a ConcurrentHashMap, and neighbors of each node are stored in Array[Int]s.
+ * Currently, only edge
+ * and node addition (not deletion) is supported.
+ */
+class ConcurrentHashMapDynamicGraph(storedGraphDir: StoredGraphDir = BothInOut)
+    extends DynamicDirectedGraphHashMap(storedGraphDir) {
+  def nodeFactory(id: Int): DynamicNode = new ConcurrentNode(id)
+  override def removeEdge(srcId: Int, destId: Int) = throw new UnsupportedOperationException()
+}
+
+private class ConcurrentNode(val id: Int) extends DynamicNode {
+  val outboundList = new ConcurrentIntArrayList()
+  val inboundList = new ConcurrentIntArrayList()
+
+  def outboundNodes(): Seq[Int] = outboundList
+  def inboundNodes(): Seq[Int] = inboundList
+  /**
+   * Add outbound edges {@code nodeIds} into the outbound list.
+   */
+  def addOutBoundNodes(nodeIds: Seq[Int]): Unit =
+    outboundList.append(nodeIds)
+
+  /**
+   * Add inbound edges {@code nodeIds} into the inbound list.
+   */
+  def addInBoundNodes(nodeIds: Seq[Int]): Unit =
+    inboundList.append(nodeIds)
+
+  def removeInBoundNode(nodeId: Int): Unit = throw new UnsupportedOperationException()
+  def removeOutBoundNode(nodeId: Int): Unit = throw new UnsupportedOperationException()
+}
+
+/** A resizable sequence of Ints which supports appending Ints (but not changing or removing
+  * current Ints).
+  * It supports concurrent reading and writing.
+  */
+// We store Ints in an array padded with extra capacity that will grow over time.
+// We essentially want a fastutil IntArrayList, but with synchronized appends.
+private class ConcurrentIntArrayList extends collection.IndexedSeq[Int] {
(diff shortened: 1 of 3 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * An efficient dynamic graph implementation which supports concurrent reading and writing.  Locks are only used by writing
+ * threads, which improves efficiency.
+ * Nodes are stored in a ConcurrentHashMap, and neighbors of each node are stored in Array[Int]s.  Currently, only edge
+ * and node addition (not deletion) is supported.
+ */
+class ConcurrentHashMapDynamicGraph(storedGraphDir: StoredGraphDir = BothInOut)
+    extends DynamicDirectedGraphHashMap(storedGraphDir) {
+  override def nodeFactory(id: Int): DynamicNode = new ConcurrentNode(id)
```
pankajgupta: shouldn't need 'override' keyword here.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
```diff
@@ -0,0 +1,106 @@
...
+  override def removeEdge(srcId: Int, destId: Int) = throw new UnsupportedOperationException()
+}
+
+private class ConcurrentNode(val id: Int) extends DynamicNode {
+  val outboundList = new ConcurrentIntArrayList()
+  val inboundList = new ConcurrentIntArrayList()
+
+  override def outboundNodes(): Seq[Int] = outboundList.toSeq
```
pankajgupta: Just to keep in mind that every access of a neighbor (e.g., every random access) will create a new class as well. If we change this to a lazy val, new updates will never be reflected -- so not sure how to solve this or if it matters?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
```diff
@@ -0,0 +1,106 @@
...
+  override def removeInBoundNode(nodeId: Int): Unit = throw new UnsupportedOperationException()
+  override def removeOutBoundNode(nodeId: Int): Unit = throw new UnsupportedOperationException()
+}
+
+/** A resizable array of Ints which supports appending Ints (but not changing or removing current Ints).
+  * It supports concurrent reading and writing.
+  */
+// We store Ints in an array padded with extra capacity that will grow over time
+// We essentially want a fastutil IntArrayList, but with synchronization.
+private class ConcurrentIntArrayList {
```
pankajgupta: how about that this just extends IndexedSeq, and have apply(idx) uses intArray(idx).

btw, also see java.util.concurrent.atomic.AtomicReference

cassovary-core/src/test/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraphSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+
+      // Test multi-edge
+      graph.addEdge(1, 2)
+      node1.inboundNodes.toList shouldEqual (List())
+      node1.outboundNodes.toList shouldEqual (List(2, 10, 2))
+      node2.inboundNodes.toList shouldEqual (List(1, 200, 1))
+      node2.outboundNodes.toList shouldEqual (List())
+    }
+
+    "support concurrent writing" in {
```
pankajgupta: ref: might want to use Futures as in SynchronizedDynamicGraphSpec, it might be easier

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
plofgren: Right, now I know the style Cassovary is using for override.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
plofgren: In high performance client code, if they want multiple random neighbors, they could still write
val outNeighbors = node.outboundNodes()
outNeighbors(5)
outNeighbors(10)
...
I would guess that most random neighbor accesses will be random walk generation, which only looks at one random neighbor from each node it walks to.

I feel somewhat nervous about lazy val because it doesn't update (as you mentioned) which could create bizarre behavior and it hides some details that could be important.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
plofgren: Nice! I think that just works, and simplifies the design.  This also avoids the object creation cost you mention above.

plofgren: Regarding AtomicReference, I considered it, but I don't think we need it here, because the assignments only happen inside a synchronized block.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
```diff
@@ -0,0 +1,106 @@
+/*
```
pankajgupta: please remove the copyright text altogether

pankajgupta:
ok, after the fixes above, this is ready to merge as well, so please just submit a squash commit including the changes

cassovary-core/src/test/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraphSpec.scala
plofgren: I remember starting to use futures for this based on SynchronizedDynamicGraphSpec, but there was some reason they were either more awkward than threads, or I felt it was harder to actually get concurrent execution.  I think the threaded code is fairly clear, given the intent is concurrent execution and not computing a result.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/ConcurrentHashMapDynamicGraph.scala
plofgren: Done

plofgren:
OK, done. Before I squashed, I also updated the scaladoc slightly.

PR #204, opened 2015-09-26
Changing to Array to save un-needed space occupied by WrappedArray

pankajgupta:
Hey @szymonm -- will appreciate if you can take a quick look.

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SharedArrayBasedDirectedNode.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SharedArrayBasedDirectedNode.scala
@@ -36,7 +36,7 @@ object SharedArrayBasedDirectedNode {
     dir match {
       case StoredGraphDir.OnlyIn | StoredGraphDir.OnlyOut | StoredGraphDir.Mutual =>
         val outEdges: Seq[Int] = Option(edges(nodeId)).getOrElse(emptyArray)
-        UniDirectionalNode(nodeId, outEdges, dir)
+        UniDirectionalNode.applySeq(nodeId, outEdges, dir)
       case StoredGraphDir.BothInOut =>
         SharedArrayBasedBiDirectionalNode(nodeId, edges, reverseDirEdges.get)
     }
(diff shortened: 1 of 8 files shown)


szymonm:
IIUC this saves creating 2 wrapped arrays per node, but even after this change we create them after calling for neighbors that returns Seq, right?

pankajgupta:
Yes, it's not stored per node though

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
```diff
@@ -112,7 +112,7 @@ class ListOfEdgesGraphReader[T](
 
       val edgesBySource = readEdgesBySource()
 
-      private lazy val edgesIterator = edgesBySource.entrySet().iterator()
+      private lazy val edgesIterator = edgesBySource.int2ObjectEntrySet().fastIterator()
```
szymonm: Did you checked what we gain by using fastIterator?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SharedArrayBasedDirectedNode.scala
```diff
@@ -36,7 +36,7 @@ object SharedArrayBasedDirectedNode {
     dir match {
       case StoredGraphDir.OnlyIn | StoredGraphDir.OnlyOut | StoredGraphDir.Mutual =>
         val outEdges: Seq[Int] = Option(edges(nodeId)).getOrElse(emptyArray)
-        UniDirectionalNode(nodeId, outEdges, dir)
+        UniDirectionalNode.applySeq(nodeId, outEdges, dir)
```
szymonm: We should have here outEdges: Array[Int] as well. But this id something for next commits. And let's wait for it until I finish my collections refactoring.

szymonm:
looks good to me

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/ListOfEdgesGraphReader.scala
pankajgupta: No, I didn't do any benchmarks but is recommended by documentation on fastutils.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/SharedArrayBasedDirectedNode.scala
pankajgupta: ok, doing nothing now then.

PR #205, opened 2015-09-28
FastMap and FastQueue implementations

### LINKED ISSUE #66: Allow node ids to be Long
This is related to https://github.com/twitter/cassovary/issues/36 but different. We want the ability of node ids to be Long. The use case is when storing a portion of a huge graph in memory on one machine. The nodes can be too many to be able to map to int. Even though the number of nodes on a single machine are small enough, they can refer to arbitrary remote nodes in their adjacency list.

This can be done by making Node and Graph to be of type[@specialized(Int, Long) V], making 'id' to be of type V and doing suitable refactoring. When the node is of type Long, a hash can be supplied to map node id to Node as required by getNodeById()

szymonm:
This commit is a first commit to resolve #66. It introduces Map and
Queue collections that are specialized, use fastutil collections
underneath and are easy to cast to Scala’s collections (in this case
they are no longer specialized).


szymonm:
Hey @pankajgupta, this is an initial commit with specialized collections.

For now, I created FastMap and FastQueue. And changed code that uses them.

The implementation of collection is a bit tricky (especially with maps and addTo) but usage is elegant - that was my goal. You should definitely first look at the usage.

Please, don't look at the details for now - I still need to document it better, make tests and benchmarks. 

Next step will be to implement FastSeq in the similar manner and use it everywhere where we store edges. This will open the way to Long ids.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+   */
+  def apply[@sp(Int, Long, AnyRef) A, @sp(Int, Long, AnyRef) B]()
+      (implicit impl: FastMapFactory[A, B, FastMap[A, B]]): FastMap[A, B] = {
+    impl()
+  }
+
+  /**
+   * @return FastMap created for given key type `A`, value type `B` and FastMap type `MapType`.
+   */
+  def applyFor[@sp(Int, Long, AnyRef) A, @sp(Int, Long, AnyRef) B, MapType <: FastMap[A, B]]()
```
szymonm: the name is confusing, but I didn't come for a better one till now.

pankajgupta:
Looking at it -- regarding naming, might want to consider CMap/CQueue where C stands for Cassovary. FastMap/Queue can be confused with fastutils and well everything in the world is fast :)

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+   * Removes key-value pair with a given key.
+   */
+  def -=(key: A): this.type
+
+  /**
+   * Get is optimized for retrieving elements contained in the Map.
+   *
+   * If the element is not in the map, returns `0` of given type (or false, null).
+   */
+  def get(key: A): B
```
pankajgupta: like Map, perhaps change this to apply(), and leave get() as returning an Option[B] ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+
+/**
+ * Specialized, fast and memory efficient implementation of mutable map.
+ *
+ * Provides basic operations only. See [[FastMapWrapper]] for how to use
+ * it as a [[mutable.Map]].
+ *
+ * Use constructor provided in the companion object.
+ */
+trait FastMap[@sp(Int, Long, AnyRef) A, @sp(Int, Long, AnyRef) B] {
```
pankajgupta: why AnyRef?

pankajgupta: may be s/B/+B/ ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+}
+
+/**
+ * Wraps `FastMap` to use as Scala's Map.
+ */
+class FastMapWrapper[A, B](underlying: FastMap[A, B]) extends mutable.Map[A, B]
+  with mutable.MapLike[A, B, FastMapWrapper[A, B]] {
+  import FastMap.Implicits._
+
+  override def +=(kv: (A, B)): FastMapWrapper.this.type = {
```
pankajgupta: do you need to specify the return type?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: If we wanted to return Option, we need to create our own specialized... This is possible, but I prefered to have implementation as fast as possible... 
There is also a problem with fastutil specialized maps and Option: they return predefined value (defaults to 0) on get, when key is not in the map. This is confusing in general setting, because we cannot distinguish it from the value that could be a value connected to the key, so to return None, we would have to call contains first...

We have similar problem with += that accepts a Tuple (in Scala), but we need to implement one ourselves to have it specialized...

pankajgupta: We do not have to create specialized for Option. We can provide both apply() and get() and add the comment about boxing in Option() and let the user of Fastmap use whichever they want.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+
+import it.unimi.dsi.fastutil.ints._
+import it.unimi.dsi.fastutil.objects._
+
+import scala.collection.generic.CanBuildFrom
+import scala.collection.mutable
+import scala.{specialized => sp}
+
+/**
+ * Specialized, fast and memory efficient implementation of mutable map.
```
pankajgupta: A couple more lines on the motivation for creating this will be helpful to the reader in the long term I think.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+
+/**
+ * Marks a [[FastMap]] with an iterator that preserves insertion order.
+ */
+trait InsertionOrderIterator
+
+/**
+ * Features creation of specialized FastMaps.
+ */
+object FastMap {
```
pankajgupta: make this extend FastMapFactory[FastMap] as this object really is a factory
where 
trait FastMapFactory[MapType[A,B] <: FastMap[A,B]] { def apply[@sp(...) A,@sp(...) B](): MapType }

Kind of like scala/collection/Map.scala and scala/collection/generic/MapFactory.scala

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
pankajgupta: why do you need both apply and applyFor ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+}
+
+/**
+ * Specialization for `Int -> Int` map.
+ */
+class FastMapIntInt[MapType <: Int2IntMap]
+  (protected val underlying: MapType)
+  extends FastMap[Int, Int] {
+
+  override def +=(key: Int, value: Int): this.type = {
```
pankajgupta: To minimize duplication of code, can some of these implementations be done in trait FastMap given underlying ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
```diff
@@ -0,0 +1,428 @@
...
+      (implicit impl: FastMapFactory[A, B, MapType]): MapType = {
+    impl()
+  }
+
+  trait FastMapFactory[@sp(Int, Long, AnyRef) A, @sp(Int, Long, AnyRef) B,
+      MapType <: FastMap[A, B]] {
+    def apply(): MapType
+  }
+
+  final class FastMapFactoryAnyAny[A, B] extends FastMapFactory[A, B, FastMap[A, B]] {
```
pankajgupta: should these be objects instead of classes?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastQueue.scala
```diff
@@ -0,0 +1,119 @@
+package com.twitter.cassovary.util.collections
+
+import it.unimi.dsi.fastutil.PriorityQueue
+import it.unimi.dsi.fastutil.ints.{IntComparator, IntHeapPriorityQueue, IntArrayFIFOQueue, IntPriorityQueue}
+import it.unimi.dsi.fastutil.objects.{ObjectArrayFIFOQueue, ObjectArrayPriorityQueue}
+
+abstract class FastQueue[@specialized(Int, Long) T] {
```
pankajgupta: Do you think we can just do FastMap as first step, then move to FastQueue in another PR. This PR is going to be big anyway.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: should be B, not +B. Map[Int, String] is not a subtype of Map[Int, Any], because put(1, new Object()) is not valid for Map[Int, String].

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: My intention was to keep API minimal and give a user guarantee that all the methods are not boxed. If a user wants get with Option, he should use asScala and then he has all the scala's map methods. This separation seems clear to me and combining boxed and pure methods in FastMap may seem confusing...
Beside that to return option we still need to check contains before... that means this method will not be fast and I wanted to avoid that.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: In a simple case I create map using just: `FastMap[Int, Int]()`
But then we have a problem, when we need addTo or preserved insertion order... Then we need to specify also the returned map type and we have `FastMap.applyFor[Int, Int, Map[Int, Int] with AddTo]()

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: It's not possible, because there is no common non-boxed superclass of Int2IntMap and Long2LongMap in fastutils. Thus, we cannot specify the type of underlying. Maybe we could do this with some weird casting, but I would rather avoid that --- I'm afraid that we will have less control over boxing then.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastQueue.scala
szymonm: I wanted to avoid fastutils in all traversers code. For this we needed Queue too. Maybe we can leave Queue here and put rest in separate commits?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: We can avoid duplication of size() and clear(), because they don't use generics (see use of JavaUtilMapWrapper) - I will fix that.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
pankajgupta: Hmmm, scala.collection.Map defines [A,+B] ...
regardless, let's leave it as B for now to reduce our worries

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
pankajgupta: Got it. ok

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
pankajgupta: may be only applyFor then?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
pankajgupta: cool.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastQueue.scala
pankajgupta: yeah, whatever you can do to break it into steps will be great. We don't have to change it all in one go

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: If you don't think it's too complicated I'll go for: applyFor renamed to apply and apply to defautl.

The usage will be:
`FastMap.default[Int, Int]()`
and
`FastMap[Int, Int, FastMap[Int, Int]]()`

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: mutable.Map is [A, B].

If you do put(1, new Object()) on immutable.Map[Int, String] you get immutable.Map[Int, Object]

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: FastMapFactory that we have needs to be parametrized with types A and B for implicits. 

object FastMapFactory is not bound to this types. We could have sth like:
```
trait FastMapFactory[MapType[A, B] <: FastMap[A, B] forSome {type A} forSome {type B}]
```
but do we really need this? Or I miss something?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastQueue.scala
szymonm: started #207

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/FastMap.scala
szymonm: objects can't be parametrized

szymonm:
Closed for now.

PR #207, opened 2015-10-01
Specialized Queue implementation

szymonm:
This is the first commit towards #66. Our goal is to wrap fastutils to
use them in Scala way. We start with Queue.

### DIFF
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
@@ -13,10 +13,9 @@
  */
 package com.twitter.cassovary.graph
 
-import it.unimi.dsi.fastutil.objects._
+import com.twitter.cassovary.util.collections.{CQueue, Order}
 import it.unimi.dsi.fastutil.ints._
-import java.util.Comparator
-
+import it.unimi.dsi.fastutil.objects._
 
 /**
  * Represents a collection of directed paths. For example, a collection of paths out of
@@ -34,7 +33,7 @@ class DirectedPathCollection {
    * Priority queue and comparator for sorting top paths. Reused across nodes.
    */
   val comparator = new PathCounterComparator(pathCountsPerId, true)
-  val priQ = new ObjectHeapPriorityQueue[DirectedPath](comparator)
+  val priQ = CQueue.priority[DirectedPath](comparator)
 
   /**
    * Appends node to the current path and record this path against the node.
@@ -68,13 +67,15 @@ class DirectedPathCollection {
 
     val pathIterator = pathCountMap.keySet.iterator
     while (pathIterator.hasNext) {
-      val path = pathIterator.next
-      priQ.enqueue(path)
+      val path = pathIterator.next()
+      priQ += path
     }
 
-    while (returnMap.size < num && !priQ.isEmpty) {
-      val path = priQ.dequeue()
+    var size = 0
+    while (size < num && !priQ.isEmpty) {
+      val path = priQ.deque()
       returnMap.put(path, pathCountMap.get(path))
+      size += 1
     }
 
     returnMap
@@ -140,7 +141,7 @@ class DirectedPathCollection {
 }
 
 class PathCounterComparator(pathCountsPerId: Int2ObjectOpenHashMap[Object2IntOpenHashMap[DirectedPath]], descending: Boolean)
-  extends Comparator[DirectedPath] {
+  extends Order[DirectedPath] {
 
   var infoMap: Object2IntOpenHashMap[DirectedPath] = null
 
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
@@ -1,0 +1,142 @@
+package com.twitter.cassovary.util.collections
+
+import it.unimi.dsi.fastutil.PriorityQueue
+import it.unimi.dsi.fastutil.ints.{IntArrayFIFOQueue, IntComparator, IntHeapPriorityQueue,
+  IntPriorityQueue}
+import it.unimi.dsi.fastutil.objects.{ObjectArrayFIFOQueue, ObjectArrayPriorityQueue}
+
+/**
+ * Specialized implementation of a queue. Queue allows adding elements and
+ * removing them in a predefined order (defined in subclasses).
+ *
+ * Implemented methods are not boxed, when using with some `Int` and `Long`.
+ */
+trait CQueue[@specialized(Int, Long) T] {
+  /**
+   * @return Next element to be dequeued.
+   */
+  def first(): T
+
+  /**
+   *
+   * @return Next element that is removed from the `CQueue`.
+   */
+  def deque(): T
+
+  /**
+   * Adds element to the `CQueue`.
+   */
+  def +=(elem: T)
+
+  def isEmpty: Boolean
+  def clear(): Unit
+  def size(): Int
+}
+
+/**
+ * FIFO [[CQueue]] - elements are added to the end of an underlying list and
+ * removed from the beginning.
+ */
+trait FIFO[@specialized(Int, Long) T] extends CQueue[T] {
+  /**
+   * Adds an element at the beginning of the underlying list.
+   */
+  def enqueueFirst(elem: T)
+}
+
+object CQueue {
+  import Implicits._
+  /**
+   * Creates specialized [[CQueue]] that is a FIFO queue.
+   */
+  def fifo[@specialized(Int, Long) T](): FIFO[T]= {
+    implicitly[CQueueFactory[T]].fifo()
+  }
+
+  /**
+   * Priority [[CQueue]] that uses default ordering (implicit) for `T`.
+   */
+  def naturalPriority[@specialized(Int, Long) T]()(implicit order: Order[T]): CQueue[T] = {
+    implicitly[CQueueFactory[T]].priority(order)
+  }
+
+  /**
+   * Priority [[CQueue]] that uses provided `order`.
+   */
+  def priority[@specialized(Int, Long) T](order: Order[T]): CQueue[T] = {
+    implicitly[CQueueFactory[T]].priority(order)
+  }
+
+  trait LowerPriorityImplicit {
+    private val factoryAny = new CQueueFactoryAny[AnyRef]()
+    implicit def factoryAny[T]: CQueueFactory[T] =
+      factoryAny.asInstanceOf[CQueueFactory[T]]
+  }
+
+  trait LowPriorityImplicit extends LowerPriorityImplicit {
+    implicit val factoryInt: CQueueFactory[Int] = new CQueueFactoryInt()
+  }
+
+  object Implicits extends LowPriorityImplicit
+
+}
+
+class CQueueAny[T, QueueType <: PriorityQueue[T]](protected val underlying: QueueType)
+    extends CQueue[T] {
+  def +=(elem: T): Unit = underlying.enqueue(elem)
+  def deque(): T = underlying.dequeue()
+  override def first(): T = underlying.first()
+  override def isEmpty: Boolean = underlying.isEmpty
+  override def clear(): Unit = underlying.clear()
+  override def size(): Int = underlying.size()
+}
+
+class CQueueInt[QueueType <: IntPriorityQueue](protected val underlying: QueueType)
+    extends CQueue[Int] {
+  override def +=(elem: Int): Unit = underlying.enqueue(elem)
+  override def deque(): Int = underlying.dequeueInt()
+  override def first(): Int = underlying.firstInt()
+  override def isEmpty: Boolean = underlying.isEmpty
+  override def clear(): Unit = underlying.clear()
+  override def size(): Int = underlying.size()
+}
+
+abstract class CQueueFactory[T] {
+  def fifo(): FIFO[T]
+  def priority(): CQueue[T]
+  def priority(order: Order[T]): CQueue[T]
+}
+
+private[collections] class CQueueFactoryAny[T] extends CQueueFactory[T] {
+  def fifo(): FIFO[T] = new CQueueAny[T, ObjectArrayFIFOQueue[T]](
+      new ObjectArrayFIFOQueue[T]()) with FIFO[T] {
+    override def enqueueFirst(elem: T): Unit = underlying.enqueueFirst(elem)
+  }
+
+  def priority(order: Order[T]): CQueue[T] = {
+      new CQueueAny[T, ObjectArrayPriorityQueue[T]](
+        new ObjectArrayPriorityQueue[T](Order.ordering(order)))
+  }
+
+  override def priority(): CQueue[T] = throw new IllegalArgumentException("Ordering missing for T")
+}
+
+private[collections] class CQueueFactoryInt extends CQueueFactory[Int] {
+  override def fifo() = new CQueueInt[IntArrayFIFOQueue](new IntArrayFIFOQueue()) with FIFO[Int] {
+    def enqueueFirst(elem: Int): Unit = underlying.enqueueFirst(elem)
+  }
+
+  def priority(order: Order[Int]): CQueue[Int] = {
+    val intComparator = new IntComparator {
+      def compare(i: Int, i1: Int): Int = order.compare(i, i1)
+
+      def compare(o1: Integer, o2: Integer): Int = {
+        throw new IllegalStateException("Boxed comparison should not be used")
+      }
+    }
+    new CQueueInt[IntHeapPriorityQueue](new IntHeapPriorityQueue(intComparator))
+  }
+
+  override def priority(): CQueue[Int] =
+    new CQueueInt[IntHeapPriorityQueue](new IntHeapPriorityQueue())
+}
--- a/cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/Order.scala
+++ b/cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/Order.scala
@@ -1,0 +1,68 @@
+package com.twitter.cassovary.util.collections
+
+/**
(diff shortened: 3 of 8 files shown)


cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/package.scala
```diff
@@ -0,0 +1,10 @@
+package com.twitter.cassovary.util
+
+import scala.language.reflectiveCalls
+
+/**
```
pankajgupta: please remove this comment, we have generally not had authors in files

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/Order.scala
```diff
@@ -0,0 +1,65 @@
...
+   */
+  def on[@specialized(Int, Long) B](f: B => A): Order[B] = new MappedOrder(this)(f)
+
+  /**
+   * Defines an ordering on `A` where all arrows switch direction.
+   */
+  def reverse: Order[A] = new ReversedOrder(this)
+}
+
+private[collections] class MappedOrder[@specialized A, @specialized B](order: Order[B])(f: A => B) extends Order[A] {
```
pankajgupta: let's continue to specialize only on Int and Long for now, rather than all primitives

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+
+import it.unimi.dsi.fastutil.PriorityQueue
+import it.unimi.dsi.fastutil.ints.{IntArrayFIFOQueue, IntComparator, IntHeapPriorityQueue,
+  IntPriorityQueue}
+import it.unimi.dsi.fastutil.objects.{ObjectArrayFIFOQueue, ObjectArrayPriorityQueue}
+
+/**
+ * Specialized implementation of queue.
+ *
+ * Implemented methods are not boxed, when using with primitive types.
```
pankajgupta: with some primitive types

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+ */
+abstract class CQueue[@specialized(Int, Long) T] {
+  def first(): T
+  def +=(elem: T)
```
pankajgupta: pls add comment on += being append

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+  def isEmpty: Boolean
+  def clear(): Unit
+  def size(): Int
+}
+
+/**
+ * FIFO [[CQueue]].
+ */
+trait FIFO[@specialized(Int, Long) T] extends CQueue[T] {
+  def enqueueFirst(elem: T)
```
pankajgupta: this is prepend, right? may be add a comment?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  override def isEmpty: Boolean = underlying.isEmpty
+
+  override def clear(): Unit = underlying.clear()
+
+  override def size(): Int = underlying.size()
+}
+
+abstract class CQueueFactory[T] {
+  def fifo(): CQueue[T] with FIFO[T]
```
pankajgupta: trait FIFO extends CQueue, so just FIFO is sufficient here?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: oh, it depends on queue implementation I guess. Sorry. Might be worth adding a comment.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+  override def priority(order: Option[Order[T]]): CQueue[T] = order match {
+    case Some(o) =>
+      new CQueueAny[T, ObjectArrayPriorityQueue[T]](
+        new ObjectArrayPriorityQueue[T](Order.ordering(o)))
+    case None =>
+      throw new IllegalArgumentException("Order must be specified for type T")
+  }
+}
+
+private[collections] class CQueueFactoryInt extends CQueueFactory[Int] {
```
pankajgupta: don't use the keyword 'override' where you don't need to

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+  override def isEmpty: Boolean = underlying.isEmpty
+
+  override def clear(): Unit = underlying.clear()
+
+  override def size(): Int = underlying.size()
+}
+
+class CQueueInt[QueueType <: IntPriorityQueue](protected val underlying: QueueType)
+    extends CQueue[Int] {
+  override def +=(elem: Int): Unit = underlying.enqueue(elem)
```
pankajgupta: Minor style comment: feel free to remove empty lines for brevity.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: may be change this to an abstract class so that you can directly instantiate it ? Or don't extend from CQueue

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+import it.unimi.dsi.fastutil.ints.{IntArrayFIFOQueue, IntComparator, IntHeapPriorityQueue,
+  IntPriorityQueue}
+import it.unimi.dsi.fastutil.objects.{ObjectArrayFIFOQueue, ObjectArrayPriorityQueue}
+
+/**
+ * Specialized implementation of queue.
+ *
+ * Implemented methods are not boxed, when using with primitive types.
+ */
+abstract class CQueue[@specialized(Int, Long) T] {
```
pankajgupta: +T ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+   * Creates specialized [[CQueue]] that is a FIFO queue.
+   */
+  def fifo[@specialized(Int, Long) T]()(implicit impl: CQueueFactory[T]): FIFO[T]= {
+    impl.fifo()
+  }
+
+  /**
+   * Priority [[CQueue]] that uses default ordering (implicit) for `T`.
+   */
+  def naturalPriority[@specialized(Int, Long) T]()(
```
pankajgupta: will the compiler allow overloaded apply() to both naturalPriority() and priority() ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  trait LowPriorityImplicit extends LowerPriorityImplicit {
+    implicit val factoryInt: CQueueFactory[Int] = new CQueueFactoryInt()
+  }
+
+  object Implicits extends LowPriorityImplicit
+
+}
+
+class CQueueAny[T, QueueType <: PriorityQueue[T]](protected val underlying: QueueType)
```
pankajgupta: s/CQueueAny/CQueueObj ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+  override def isEmpty: Boolean = underlying.isEmpty
+
+  override def clear(): Unit = underlying.clear()
+
+  override def size(): Int = underlying.size()
+}
+
+abstract class CQueueFactory[T] {
+  def fifo(): CQueue[T] with FIFO[T]
+  def priority(order: Option[Order[T]]): CQueue[T]
```
pankajgupta: how about not having Option, but having two overloaded methods:
1. def priority()
2. def priority(order: Order[T])
Just makes it clearer and there is no duplication of code because you have the two cases of Some() and None coded anyway

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * FIFO [[CQueue]].
+ */
+trait FIFO[@specialized(Int, Long) T] extends CQueue[T] {
+  def enqueueFirst(elem: T)
+}
+
+object CQueue {
```
pankajgupta: I think if this extends CQueueFactory, you can just delegate to it rather than using impls. Like how object Map extends MapFactory (in scala.collection.Map)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
```diff
@@ -13,10 +13,11 @@
  */
 package com.twitter.cassovary.graph
 
-import it.unimi.dsi.fastutil.objects._
+import com.twitter.cassovary.util.collections.{CQueue, Implicits, Order}
 import it.unimi.dsi.fastutil.ints._
-import java.util.Comparator
+import it.unimi.dsi.fastutil.objects._
 
+import Implicits._
```
pankajgupta: it's not that great to have to import Implicits._ everytime we have to use CQueue.... Any solution?

cassovary-core/src/test/scala/com/twitter/cassovary/util/collections/FastQueueSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.NoSuchElementException
+
+import com.twitter.cassovary.util.collections.Implicits._
+import org.scalatest.{Matchers, WordSpec}
+
+/**
+ * Created by szymonmatejczyk on 23.09.15.
+ */
+class FastQueueSpec extends WordSpec with Matchers {
```
pankajgupta: s/Fast/C everywhere

cassovary-core/src/test/scala/com/twitter/cassovary/util/collections/FastQueueSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+  "FastQueue" should {
+    "enqueue/deque elements" when {
+      "using fifo" in {
+        val q = CQueue.fifo[Int]()
+        q.isEmpty should be (true)
```
pankajgupta: refactor common code in these test examples

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: I don't really get what you mean. My thoughts are:
* We instantiate only in the companion object to get implicits working. 
* The only difference I know of between ABC and trait is that trait has no fields, so can be mixed in. Here, trait is ok since FIFO has no fields. 
*It should extend CQueue because FIFO is always a Queue.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: That would mean that Queue[Int] is a Queue[Any] - not true. Same here: http://www.scala-lang.org/api/current/index.html#scala.collection.mutable.Queue

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: No, tried this. It comlains that after type erasure methods are the same... The method names must be different. Maybe you have better ideas?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: Any is a scala name for Object, but if you like Obj - can change it.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: The problem is we have to pass Order implicitly for 1. and then compiler complains about same method types after erasure.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: Factories are bind to specific type. We need this type for implicits resolution --- to inject code depeneding on a generic type.

object CQueue is not bind to a specific type - just methods are parametrized.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/DirectedPathCollection.scala
szymonm: Done this.

szymonm:
Check failure seems not to be my fault, can you @pankajgupta have a look at it?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: no Any is fine too

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
```diff
@@ -0,0 +1,142 @@
...
+ */
+trait FIFO[@specialized(Int, Long) T] extends CQueue[T] {
+  /**
+   * Adds an element at the beginning of the underlying list.
+   */
+  def enqueueFirst(elem: T)
+}
+
+object CQueue {
+  import Implicits._
```
pankajgupta: may be just define all those factory implicits here here so that there is no need for traits and object Implicits?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: bad langauge on my part -- by "directly instantiate" I actually meant not having to use the more verbose "CQueue with FIFO" everywhere.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: actually, priority() should not be there at all. Usually if there is a method with an implicit param, a caller can always specify an explicit param

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: ok

pankajgupta:
About the failing check, it is really odd as it fails for just one travis config (and is okay in master). Will look more if this continues

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: We need the traits for implicits precedense. Basically: if you can CQueue[Int]() you are matching both factories: CQueueFactory[T] called factoryAny from LowerPriorityImplicit and CQueueFactory[Int] called factoryInt from LowPriorityImplicit. 
Scala will choose the latter one, because it's defined in the subclass of the class where the former is defined. Thus, I'm afraid we can't get rid of this nasty traits...
Alternatively, we can make CQueue extends LowPriorityImplicits but this seems not intuitive for me.

szymonm:
Seems like we are done but the import Implicits._. 
I will squash the commits to pass travis...

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/package.scala
```diff
@@ -0,0 +1,5 @@
+package com.twitter.cassovary.util
+
+package object collections {
+  object Implicits extends CQueue.LowPriorityImplicit
```
pankajgupta: needed any more?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
pankajgupta: btw, another alternative to implicits is using typeTags. not saying it is better, but something to consider:
A toy example to do something specific for Int:

import scala.reflect.runtime.universe._
class Q[T] { }
abstract class QFac[T] {def fifo(): Q[T] }

val qFacAny = new QFac[AnyRef] { def fifo() = new Q[AnyRef] }
val qFacInt = new QFac[Int] { def fifo() = { println("return anything special for ints") ; new Q[Int] } }
object Q {
def fifo[T: TypeTag] = { if (typeTag[T] == typeTag[Int]) qFacInt.fifo() else qFacAny.fifo() }
}

szymonm:
ball in your court, @pankajgupta

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CQueue.scala
szymonm: yes, this seems possible too, but TypeTag uses reflection that can be slower (if we ever care about creation time...)

PR #211, opened 2016-01-26
#200 Feature/fastseq

### LINKED ISSUE #200: Convert Node.inboundNodes() and outboundNodes() to have Array[Int] instead of Seq[Int]
I think @szymonm you had pointed this out with some benchmarking data as well? If so, can you please link to a gist if possible. I am finally coming around to agreeing that this should indeed be done as the current code does un-necessary boxing/unboxing.

szymonm:
Hi @pankajgupta,

implemented specialized CSeq as we discussed. Can you have a look at it?

I will be working on additional tests for CSeq and some documentation in the next days.

I decided to force use to import implicits manually, because I had some troubles with importing them only in factory creating CSeq. It turned out that Scala didn't use specialized versions of my collections. To avoid this kind of bugs for CSeq I'm not providing `CSeqFactory[Any]`. I don't think we should need one as we want CSeq be used only for Int and Long for now.

I also wanted to keep CSeq interface minimal but needed `slice` for efficiency and type consistency of methods such as `Node.inboundNodes(max: Int)`. However, I also consider a breaking change of deleting all the methods with `max` number of elements retrieved since a user can use `.toSeq.take(max)`.

I also need to write some benchmark to check the result of this change on the performance.

Resolves #200


szymonm:
Hey @pankajgupta, ball in your court here.

pankajgupta:
yes, pardon the delay. getting to it.

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CSeq.scala
```diff
@@ -0,0 +1,158 @@
+package com.twitter.cassovary.util.collections
+
+import scala.collection.immutable
+
+trait CSeq[@specialized(Int, Long) T] {
```
pankajgupta: pls add some comments for the reason of existence (brief summary of what we discussed I guess)

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CSeq.scala
```diff
@@ -0,0 +1,158 @@
...
+  class IntAndLongImplicits {
+    val cSeqFactoryInt: CSeqFactory[Int] = new CSeqFactoryInt
+    implicit def csi: CSeqFactory[Int] = cSeqFactoryInt
+    implicit val cSeqFactoryLong: CSeqFactory[Long] = new CSeqFactoryLong
+  }
+
+  object Implicits extends IntAndLongImplicits
+}
+
+abstract class CSeqFactory[@specialized(Int, Long) T] {
```
pankajgupta: separate into a different file

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CSeq.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  def fromCSeq(arr: CSeq[T], from: Int, until: Int): CSeq[T]
+
+  def fromArray(arr: Array[T]): CSeq[T]
+
+  def empty: CSeq[T]
+}
+
+
+class CSeqFactoryInt extends CSeqFactory[Int] {
```
pankajgupta: pls see http://docs.scala-lang.org/overviews/macros/implicits for perhaps using implicit macros to avoid this boilerplate between Int and Long ?

cassovary-core/src/main/scala/com/twitter/cassovary/util/collections/CSeq.scala
szymonm: When using Macros this wasn't that easy. I needed CSeq, CSeqFactory and CSeqWrapper in the same file to be visible by the macro.

szymonm:
looks like 2.10 compatibility is a bit hard...

pankajgupta:
We can sunset 2.10

szymonm:
Managed to make it 2.10 compliant.

Your turn, @pankajgupta

szymonm:
Checked running time on global PageRank benchmark and it is the same.

pankajgupta:
Cool. Is there any benchmark which highlights the benefit of the change --
i.e., the original motivation of doing it?

szymonm:
Main goal of this change was to allow easier switch to Long ids by unifying underlying collections.

Second goal was to avoid boxing where possible. I think we should see an improvement in PageRank after I change its implementation to avoid using toSeq method on CSeq - will check that in the next days.

cassovary-collections/README.md
```diff
@@ -0,0 +1,9 @@
+# Cassovary-collections
+
+This subproject contains specialized collections that are used by Cassovary library.
+
+General goals for the collections implemented in this subproject are:
+
+1) low memory footprint (storing primitive types where possible);
```
pankajgupta: remove semicolons

cassovary-collections/README.md
```diff
@@ -0,0 +1,9 @@
+# Cassovary-collections
+
+This subproject contains specialized collections that are used by Cassovary library.
+
+General goals for the collections implemented in this subproject are:
+
+1) low memory footprint (storing primitive types where possible); 
+2) optimization for Int and Long types (no boxing);
```
pankajgupta: may be refer to a benchmark file specifically comparing CSeq to IndexedSeq for ints / longs ?

szymonm:
Ran improved PageRank vs the previous one:

| graph     | before                                   | improved                                         |
|-----------|------------------------------------------|--------------------------------------------------|
| facebook |  40ms  | 24ms |
| wiki-Vote |             55ms                             |      35ms                                            |

pankajgupta:
@szymonm just confirming that this is the PR to be reviewed?

szymonm:
@pankajgupta yes, this one.

cassovary-benchmarks/BenchmarksREADME.md
```diff
@@ -3,10 +3,21 @@
...
 In the main project folder type `./sbt` to enter sbt interactive mode. Move to
 cassovary-benchmarks project using
 `project cassovary-benchmarks`. Now by typing `run` you can see and choose one
 of the benchmarks you would like to run. You can also
 use `runMain BenchmarkClassName parameters` to run a particular benchmark,
 for example `runMain com.twitter.cassovary.PerformanceBenchmark -local=facebook -globalpr`
+
+## Building and running - JMH benchmarks
+
+JMH is a JVM benchmarking tool that makes sure that your code is tested after JIT
```
pankajgupta: good to know about JMH

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
```diff
@@ -93,73 +93,86 @@ object PerformanceBenchmark extends App with GzipGraphDownloader {
...
-        AdjacencyListGraphReader.forIntIds(path, filename).toArrayBasedDirectedGraph()
-      } else {
-        val sep = separatorInt().toChar
-        printf("Using Character (%d in Int) as separator\n", sep.toInt)
-        ListOfEdgesGraphReader.forIntIds(path, filename, graphDir = StoredGraphDir.BothInOut,
-          separator = sep).toSharedArrayBasedDirectedGraph(forceSparseRepr = None)
-          //separator = sep).toArrayBasedDirectedGraph(neighborsSortingStrategy = LeaveUnsorted,
-          //      forceSparseRepr = None)
-      }
+  def main(args: Array[String]): Unit = {
```
pankajgupta: any particular reason for this refactoring into main() ?

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/utils/CSeqVsSeqBenchmark.scala
```diff
@@ -0,0 +1,48 @@
...
+ */
+package com.twitter.cassovary.utils
+
+
+import com.twitter.cassovary.collections.CSeq
+import org.openjdk.jmh.annotations.{Benchmark, Scope, Setup, State}
+import CSeq.Implicits._
+
+@State(Scope.Thread)
+class CSeqVsSeqBenchmark {
```
pankajgupta: perhaps change the name to CSeqVsSeqIntBenchmark ?

cassovary-collections/README.md
```diff
@@ -0,0 +1,10 @@
+# Cassovary-collections
+
+This subproject contains specialized collections that are used by Cassovary library.
+
+General goals for the collections implemented in this subproject are:
+
+1) low memory footprint (storing primitive types where possible)
+2) optimization for Int and Long types (no boxing) (see `CSeqVsSeq` benchmark in 
+`cassovary-benchmarks`)
+3) generic implementation and ease of convertion to Scala's collections.
```
pankajgupta: s/convertion/conversion

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  def fromCSeq(arr: CSeq[T], from: Int, until: Int): CSeq[T]
+
+  def fromArray(arr: Array[T]): CSeq[T]
+
+  def empty: CSeq[T]
+}
+
+/**
+  * Companion object that should be used, when creating `CSeq`s.
```
pankajgupta: Please change comment to "Companion object to CSeq. This object's apply methods should be used to create new `CSeq`s."

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  def empty[@specialized(Int, Long) T](implicit cSeqFactory: CSeqFactory[T]) = {
+    cSeqFactory.empty
+  }
+
+  object Implicits {
+    private[this] val counter = new AtomicInteger(0)
+
+    // We use fundep materialization to generate classes for specific types
```
pankajgupta: can you please add a comment here that we are using macros to avoid boilerplate. Otherwise we need to write multiple instance definitions of CSeqFactory

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
```diff
@@ -0,0 +1,166 @@
...
+    implicit def materializeCSeqFactory[T]: CSeqFactory[T] = macro impl[T]
+    def impl[T: c.WeakTypeTag](c: scala.reflect.macros.Context): c.Expr[CSeqFactory[T]] = {
+      import c.universe._
+
+      // In Scala 2.10 we need to manually ensure that type names are not duplicated
```
pankajgupta: just wondering why type names would be duplicated as className includes sym.toString ?

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
```diff
@@ -0,0 +1,166 @@
...
+              throw new IndexOutOfBoundsException(idx.toString)
+            } else {
+              underlying.apply(from + idx)
+            }
+          }
+
+          override def length: Int = until - from
+        }
+
+        class $arrayWrapperName(underlying: Array[$sym])
```
pankajgupta: do we need this extra class, or can we just make a function here to create a CSeq, i.e.,
def fromArray(underlying: Array[$sym]) = new CSeq[$sym] { ... }

and use fromArray below instead of $arrayWrapperName ?

pankajgupta: It is okay either way btw from my side if you prefer to have a class and not directly created traits

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/SortedCSeqOps.scala
```diff
@@ -19,7 +33,8 @@ object SortedArrayOps {
   /**
    * Linear intersection of two sorted arrays.
    */
-  def intersectSorted(array: Array[Int], that: Array[Int]): Array[Int] = {
+  def intersectSorted(array: CSeq[Int], that: CSeq[Int])(implicit csf: CSeqFactory[Int]):
```
pankajgupta: s/array/cseq1 may be?

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/SortedCSeqOps.scala
```diff
@@ -80,13 +96,8 @@ object SortedArrayOps {
...
     }
 
-    slice(union, 0, resultIdx)
-  }
-
-  @inline def slice(a: Array[Int], from: Int, len: Int): Array[Int] = {
-    val result = Array.ofDim[Int](len)
-    Array.copy(a, from, result, 0, len)
-    result
+    // possible memory leak
```
pankajgupta: why a memory leak possibly?

pankajgupta:
Just fyi, I am in the middle of my review and might take 1-2 days. Will write here when I am finished with all of it.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/utils/CSeqVsSeqBenchmark.scala
```diff
@@ -0,0 +1,48 @@
...
+  val seqsLength = 100000
+
+  var seq :Seq[Int] = _
```
pankajgupta: minor, but for the sake of consistency, space should not be before the colon, but after.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/utils/CSeqVsSeqBenchmark.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  var cSeq :CSeq[Int] = _
+
+  @Setup
+  def prepareSeqs() = {
+    val arr = Array.tabulate(seqsLength)(n => n)
+    seq = arr.toSeq
+    cSeq = CSeq[Int](arr)
```
pankajgupta: perhaps just CSeq(arr) works?

cassovary-collections/src/test/scala/com/twitter/cassovary/collections/CSeqSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+  "CSeq" should {
+    "behave like an array" when {
+      "created for int" in {
+        val s = Array.ofDim[Int](5)
+        s(0) = 0
+        s(1) = 1
+
+        val cSeq = CSeq(s)
+
+        cSeq.length shouldEqual (5)
```
pankajgupta: don't need parentheses around 5, btw

cassovary-collections/src/test/scala/com/twitter/cassovary/collections/CSeqSpec.scala
```diff
@@ -0,0 +1,50 @@
...
+
+      "comparing CSeq[Int] with CSeq[Long]" in {
+        val s1 = CSeq[Int](Array(1, 2, 3))
+        val s2 = CSeq[Long](Array[Long](1, 2, 3))
+
+        s1 shouldEqual (s2)
+      }
+    }
+  }
+}
```
pankajgupta: may be test slice and foreach as well?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -13,8 +13,12 @@
  */
 package com.twitter.cassovary.graph
 
+import com.twitter.cassovary.collections.CSeq
+
 import scala.util.Random
 
+import com.twitter.cassovary.collections.CSeq.Implicits._
```
pankajgupta: sort in alpha order

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
```diff
@@ -0,0 +1,166 @@
...
+      // val className = internal.reificationSupport.freshTypeName("CSeq" + sym.toString + "$")
+      // val arrayWrapperName = internal.reificationSupport.freshTypeName("CSeq" + sym.toSt ring +
+      //        "ArrayWrapper$")
+      // val factoryClassName = internal.reificationSupport.freshTypeName("CFactory" + sym
+      // .toString + "$")
+      val clazz = q"""
+        class $className(underlying: CSeq[$sym], from: Int, until: Int)
+            extends com.twitter.cassovary.collections.CSeq[$sym] {
+          override def apply(idx: Int) = {
+            if (idx >= length) {
```
pankajgupta: do we need to check this you think, or underlying.apply() will check it anyway? (One problem with leaving it to underlying.apply() might be that the ArrayIndexOutOfBoundsException might give the wrong index...)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -217,12 +221,12 @@ trait Node {
...
    */
-  def intersect(dir: GraphDir, nodeIds: Seq[Int]): Seq[Int] = {
+  def intersect(dir: GraphDir, nodeIds: CSeq[Int]): CSeq[Int] = {
     intersect(neighborIds(dir), nodeIds)
   }
 
-  protected def intersect(neighbors: Seq[Int], nodeIds: Seq[Int]): Seq[Int] = {
-    neighbors.intersect(nodeIds)
+  protected def intersect(neighbors: CSeq[Int], nodeIds: CSeq[Int]): CSeq[Int] = {
+    CSeq[Int](neighbors.toSeq.intersect(nodeIds.toSeq).toArray)
```
pankajgupta: can we have a special implementation of intersect please for CSeqs? Leave this method around for when supplied nodeIds are of type Seq[Int]

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/BiDirectionalNode.scala
```diff
@@ -29,15 +32,15 @@ class FillingInEdgesBiDirectionalNode(val id: Int, val outEdges: Array[Int])
 
   var inEdges: Array[Int] = BiDirectionalNode.noEdges
 
-  def inboundNodes(): Seq[Int] = inEdges
+  def inboundNodes(): CSeq[Int] = CSeq(inEdges)
 
-  def outboundNodes(): Seq[Int] = outEdges
+  def outboundNodes(): CSeq[Int] = CSeq[Int](outEdges)
```
pankajgupta: for consistency, do not have CSeq[Int](outEdges) just CSeq(outEdges)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
```diff
@@ -335,7 +335,7 @@ trait QueueBasedTraverser[+V <: Node] extends Traverser[V] {
       degreeLimit.isLimitReached(node.neighborCount(dir))) {
       Seq()
     } else {
-      limitAddedToQueue(node.neighborIds(dir).filter(n => shouldBeEnqueued(coloring.get(n))))
+      limitAddedToQueue(node.neighborIds(dir).toSeq.filter(n => shouldBeEnqueued(coloring.get(n))))
```
pankajgupta: can we implement filter and map efficiently as well in CSeq?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
```diff
@@ -15,10 +15,13 @@ package com.twitter.cassovary.graph.bipartite
 
 import com.twitter.cassovary.graph.{Graph, GraphDir, StoredGraphDir, Node}
 import com.twitter.cassovary.graph.GraphDir._
+import com.twitter.cassovary.collections.CSeq
 import com.twitter.logging.Logger
 
 case class BipartiteGraphException(e: String) extends Exception(e)
 
+import com.twitter.cassovary.collections.CSeq.Implicits._
```
pankajgupta: move up to the other imports

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/BiDirectionalNode.scala
```diff
@@ -29,15 +32,15 @@ class FillingInEdgesBiDirectionalNode(val id: Int, val outEdges: Array[Int])
 
   var inEdges: Array[Int] = BiDirectionalNode.noEdges
 
-  def inboundNodes(): Seq[Int] = inEdges
+  def inboundNodes(): CSeq[Int] = CSeq(inEdges)
```
pankajgupta: CSeq[Int](inEdges) ?

project/Build.scala
```diff
@@ -1,8 +1,10 @@
+package com.twitter.cassovary.colections
+
 import sbt._
 import Keys._
 import sbtassembly.AssemblyPlugin.autoImport._
 import xerial.sbt.Sonatype._
-
+import pl.project13.scala.sbt.JmhPlugin
 
 object Cassovary extends Build {
```
pankajgupta: change CassovaryLibraryVersion to 7.0.0

pankajgupta:
Finished review. Looks good -- just some minor comments.

Please check why travis-ci is failing. If it helps, I still say, let's drop support for scala 2.10 and move on to 2.11 only.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
szymonm: Sbt JMH plugin gets crazy and fails just after start when using App trait, because of using DelayedInit.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/utils/CSeqVsSeqBenchmark.scala
szymonm: done.

cassovary-collections/README.md
szymonm: done.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
szymonm: done.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
szymonm: Don't know exactly why, but the macro code is called whenever we create CSeq[Int]. This seams to be normal in fundep materialization.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
szymonm: In this case we must create this class anonymously in 3 places: ofDim, empty and fromArray, which can be confusing.

Let me leave it this way.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/SortedCSeqOps.scala
szymonm: right

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/SortedCSeqOps.scala
szymonm: The CSeq we create in the next line is a slice of array defined in line 34 that has length: min(array.length, that.length). Cells from intersectionIdx to the end are wasted. 

Alternatively we can use ArrayBuffer from fastutils (that is not ready yet) or copy indices before intersectionIdx to a new array (which should be equivalent). In this case we waste speed to gain some memory. What do you think about this tradeoff?

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/utils/CSeqVsSeqBenchmark.scala
szymonm: done.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
szymonm: done.

cassovary-collections/src/test/scala/com/twitter/cassovary/collections/CSeqSpec.scala
szymonm: done.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
szymonm: Yes, and I don't want to allow a slice read/write outside it's scope.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
szymonm: I don't think I understood. Why would we need intersect for Seq[Int] now as it's protected and we want to use CSeq inside cassovary.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
```diff
@@ -242,15 +246,15 @@ trait Node {
    * more optimized implementation.
    * @return a boolean indicating Whether `nodeIds` contains `queryNodeId`.
    */
-  protected def containsNode(nodeIds: Seq[Int], queryNodeId: Int): Boolean = {
-    nodeIds.contains(queryNodeId)
+  protected def containsNode(nodeIds: CSeq[Int], queryNodeId: Int): Boolean = {
+    nodeIds.toSeq.contains(queryNodeId)
```
szymonm: sure, done.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: Not really
1) filter
As we don't have ArrayBuffer, we would have to copy filtered elements at least twice.
2) map
if we wanted CSeq[Int].map[T](f: Int => T): CSeq[T]
  If we don't have CSeq[T] that is efficiently implemented we loose the efficiency.
if we wanted CSeq[Int].map(f: Int => Int): CSeq[Int] we are not consistent with scala's map API and this version doesn't give as much.

szymonm: what do you think?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/bipartite/BipartiteGraph.scala
szymonm: done.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/BiDirectionalNode.scala
szymonm: Can you explain a bit more?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/BiDirectionalNode.scala
szymonm: done

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
pankajgupta: is that only in 2.10? so we are creating a new class for every time we create CSeq?

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
pankajgupta: sure

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/SortedCSeqOps.scala
pankajgupta: probably copy to new array (also add a comment about this tradeoff / memory leak)

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
pankajgupta: I meant implement intersect for CSeq not using conversion to Seq, but write new code for native intersection of CSeqs.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: 1. filter: dunno which will be more expensive: (a) converting to Seq, or (b) counting the number of filtered items and then allocating an array for it and putting the items in.
2. (A) If `CSeq[T]` is efficiently implemented (Int or Long), will it be okay? If so, we are good?

cassovary-core/src/main/scala/com/twitter/cassovary/graph/node/BiDirectionalNode.scala
pankajgupta: don't remember now. please ignore...

pankajgupta:
ping @szymonm

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
szymonm: Done.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
szymonm: It happens in both 2.10 and 2.11. We are create a class (in compile time) for every time CSeq is created.

szymonm: As far as I know, but I don't understand the macros very deeply.

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/SortedCSeqOps.scala
szymonm: Done.

cassovary-benchmarks/src/main/scala/com/twitter/cassovary/utils/CSeqVsSeqBenchmark.scala
szymonm: Done.

cassovary-collections/src/test/scala/com/twitter/cassovary/collections/CSeqSpec.scala
szymonm: done

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
szymonm: To do this efficiently, we need specialized Map[T, Int]. Let's leave it for now.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
szymonm: 1. That depends on the filter function itself I think... 

Both possible, but how about leaving it for a separate issue?

szymonm:
Sorry for the delay. Your turn, @pankajgupta

cassovary-collections/src/main/scala/com/twitter/cassovary/collections/CSeq.scala
pankajgupta: Can you verify that this class is not being created at runtime as well every time CSeq is created.

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Node.scala
pankajgupta: ok

cassovary-core/src/main/scala/com/twitter/cassovary/graph/Traverser.scala
pankajgupta: sure

pankajgupta:
ready to merge once you are able to double check that no new class is being created at runtime as well (as mentioned above)

szymonm:
Do you have any paticular idea on how to check that the classes are not generated in runtime?

I checked what is generated in compile time and the looks like the classes are generated then:
```
➜  cassovary git:(feature/fastseeq) ls cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node\$CSeq*
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqInt$4$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqInt$5$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqInt$6$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqInt$7$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqIntArrayWrapper$4$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqIntArrayWrapper$5$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqIntArrayWrapper$6$1.class
cassovary-core/target/scala-2.11/classes/com/twitter/cassovary/graph/Node$CSeqIntArrayWrapper$7$1.class
```

szymonm:
I have no idea why travis is failing for tests that seem to be unrelated. Looks like it is some Scala's compiler internal issue. Let's drop support for 2.10.

pankajgupta:
ok to stop supporting 2.10 @szymonm -- is some code only for 2.10? if so, please clean that up.

also, perhaps try writing some print statements to test whether the class is being generated at runtime or not?

szymonm:
yes, there is some code for 2.10. I'm cleaning it right now. And I'll squash all the commits.

I tested with print statements and they are run only in compile time.

pankajgupta:
awesome

PR #221, opened 2016-08-24
Bash example

yb1:
Hi,

It seemed that there were some trivial errors (e.g trying to load graph 5M before generating it with WriteRandomGraph. I switched the order of generating 5M and 500K ones), so I made little modifications.

### DIFF
--- a/cassovary-examples/src/main/bash/load-graph-examples.sh
+++ b/cassovary-examples/src/main/bash/load-graph-examples.sh
@@ -17,14 +17,14 @@ JARDIR=cassovary-examples/target/scala-2.11
 JARS=`echo $JARDIR/*assembly*.jar`
 JARSCP=${JARS// /:}
 
-echo Generating a random graph with 500K nodes with average outdegree 20 (so ~10M edges total in one direction) and storing in /tmp/...
-java -Xmx1200M -server -cp $JARSCP WriteRandomGraph /tmp 5M 5000000 20 8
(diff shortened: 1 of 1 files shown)


cassovary-examples/src/main/bash/load-graph-examples.sh
```diff
@@ -17,14 +17,14 @@ JARDIR=cassovary-examples/target/scala-2.11
...
 echoandrun 120M 500K BothInOut
 
-echo Generating a random graph with 5M nodes with average outdegree 20 (so ~100M edges total in one direction) and storing in /tmp/...
-java -Xmx1500M -server -cp $JARSCP WriteRandomGraph /tmp 500K 500000 20 4
+echo "Generating a random graph with 5M nodes with average outdegree 20 (so ~100M edges total in one direction) and storing in /tmp/..."
+java -Xmx1500M -server -cp $JARSCP WriteRandomGraph /tmp 5M 5000000 20 8
 
 echoandrun 500M 5M OnlyOut
-echoandrun 1100M 5M BothInOut
+echoandrun 1200M 5M BothInOut
```
pankajgupta: Rest looks good but did it not run for you with 1100MB memory?

yb1: I thought it didn't (saw memory problem). I just tried few more times with 1100M, and they were all successful. So I modified it back to 1100M. Thanks.

Issue #223, opened 2016-10-09
Add optional command-line arg in PerformanceBenchmark to store only outgoing edges


lintool:
Current implementation of `PerformanceBenchmark` is fixed to read graph as `StoredGraphDir.BothInOut`. Let's add a command-line option to store only outgoing edges, `StoredGraphDir.OnlyOut`

Ref: https://github.com/twitter/cassovary/blob/master/cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala#L112

lintool:
This issues has been addressed.

Issue #222, opened 2016-10-09
PerformanceBenchmark doesn't seem to be able to read gzipped files


lintool:
The class `com.twitter.cassovary.PerformanceBenchmark` doesn't seem to be able to read gzipped files. Example:

```
[jimmylin@tibanna cassovary]$ ls cassovary-benchmarks/src/main/resources/graphs/soc-LiveJournal1.txt.gz 
cassovary-benchmarks/src/main/resources/graphs/soc-LiveJournal1.txt.gz
[jimmylin@tibanna cassovary]$ ./sbt "project cassovary-benchmarks" "run-main com.twitter.cassovary.PerformanceBenchmark -local=soc-LiveJournal1 -separator=9 -globalpr"
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option PermSize=256m; support was removed in 8.0
[info] Loading project definition from /scratch0/jimmylin/cassovary/project
[info] Set current project to cassovary (in build file:/scratch0/jimmylin/cassovary/)
[info] Set current project to cassovary-benchmarks (in build file:/scratch0/jimmylin/cassovary/)
[info] Running com.twitter.cassovary.PerformanceBenchmark -local=soc-LiveJournal1 -separator=9 -globalpr
Reading soc-LiveJournal1 graph from src/main/resources/graphs
Using Character (9 in Int) as separator
Oct 08, 2016 11:04:03 PM com.twitter.cassovary.util.io.FileReader <init>
INFO: Starting reading from file src/main/resources/graphs/soc-LiveJournal1.txt.gz...

Exception in thread "main" java.nio.charset.MalformedInputException: Input length = 1
    at java.nio.charset.CoderResult.throwException(CoderResult.java:281)
    at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:339)
    at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
    at java.io.InputStreamReader.read(InputStreamReader.java:184)
    at java.io.BufferedReader.fill(BufferedReader.java:161)
    at java.io.BufferedReader.readLine(BufferedReader.java:324)
    at java.io.BufferedReader.readLine(BufferedReader.java:389)
    at scala.io.BufferedSource$BufferedLineIterator.hasNext(BufferedSource.scala:72)
    at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:408)
    at com.twitter.cassovary.util.io.FileReader.checkNext(FileReader.scala:40)
[...]

PR #224, opened 2016-10-19
GraphReaders read Gzip files

yb1:
Hi,
I added functionality for GraphReaders' to directly read from gzip file 
Also few minor flags for benchmark were added for flexibility.
Related to issues   #222 and #223


cassovary-benchmarks/src/main/scala/com/twitter/cassovary/PerformanceBenchmark.scala
```diff
@@ -93,6 +93,10 @@ object PerformanceBenchmark extends GzipGraphDownloader {
   val getNodeFlag = flags("gn", 0, "run getNodeById benchmark with a given number of steps")
   val reps = flags("reps", DEFAULT_REPS, "number of times to run benchmark")
   val adjacencyList = flags("a", false, "graph in adjacency list format")
+  val gzipFlag = flags("gz", false, "Is file in gzip format")
+  val dirOutFlag = flags("out", false, "Graph direction: out")
```
pankajgupta: I think you want the default values to be true for both dirOutFlag and dirInFlag so that graphDir = BothInOut by default

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/FileReader.scala
```diff
@@ -6,15 +6,24 @@ import com.twitter.logging.Logger
 import com.twitter.util.NonFatal
 
 import scala.io.Source
+import java.io.FileInputStream
```
pankajgupta: to maintain alphabetic order of the imports, please put these new imports above scala.io.Source

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReader.scala
```diff
@@ -13,11 +13,13 @@
...
-import com.twitter.cassovary.graph.{StoredGraphDir, NodeIdEdgesMaxId}
+import com.twitter.cassovary.graph.{NodeIdEdgesMaxId, StoredGraphDir}
 import com.twitter.cassovary.graph.StoredGraphDir._
-import com.twitter.cassovary.util.{ParseString, NodeNumberer}
+import com.twitter.cassovary.util.{NodeNumberer, ParseString}
 import com.twitter.util.NonFatal
-import java.io.IOException
+import java.io.{BufferedInputStream, FileInputStream, IOException}
+import java.util.zip.GZIPInputStream
+
```
pankajgupta: delete empty line

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/AdjacencyListGraphReader.scala
```diff
@@ -80,7 +83,10 @@ class AdjacencyListGraphReader[T] (
     override def iterator = new Iterator[NodeIdEdgesMaxId] {
 
       var lastLineParsed = 0
-      private val src = Source.fromFile(filename)
+      private val src = if (!isGzip)
```
pankajgupta: this is replicated code from FileReader. It was already there before your changes but now this extra code is making it worse. Can we instantiate a FileReader just like we instantiate TwoTsFileReader in ListOfEdgesGraphReader?

cassovary-core/src/main/scala/com/twitter/cassovary/util/io/FileReader.scala
```diff
@@ -85,4 +85,24 @@ class TwoTsFileReader[T](fileName: String,
     (source, dest)
   }
 
+}
+
+// T is the id type, typically Int or Long or String
+class AdjacencyTsFileReader[T](fileName: String,
```
pankajgupta: Instead of having "Any", you should change it (T, Option[Int]) -- where T is either the source or dest id and "Int" represents the count of adjacency.

Also please add a comment about this.

yb1:
I've just fixed it. Thanks, Pankaj!

